PR #35, opened 2024-07-01
fix(): adjust interrupt logic

sunshinexcode:


cyfyifanchen:
Works on all languages?

agents/addon/extension/interrupt_detector/extension.go
```diff
@@ -51,10 +51,9 @@ func (p *interruptDetectorExtension) OnData(
...
 	}
 
-	final, err := data.GetPropertyBool(textDataFinalField)
-	if err != nil {
-		slog.Warn(fmt.Sprintf("OnData GetProperty %s error: %v", textDataFinalField, err), logTag)
-		return
+	var final bool
+	if len(text) >= 2 {
+		final = true
 	}
```
wangyoucao577, line 57: It might be better to use `if final || len(text) >= 2` instead?

wangyoucao577 (approved):

wangyoucao577:
> Works on all languages?

Good point!

Issue #81, opened 2024-07-10
[Feature Request] Thinking about using Docker compose at the root of the project. 


cyfyifanchen:
`docker-compose.yml` has it's pros and cons. 

Someone in the community suggested that using `docker-compose` at the project root would be more convenient, allowing the environment to be set up with a simple `docker compose up -d`. They listed the benefits and, if consensus is reached, this approach could be considered for inclusion in the project root.

Lensual:
对于想快速体验的开发者来说，docker compose来代替docker run会方便一些
1. 不依赖本地编译环境 node、go
  - 解决本地环境版本不一致
  - 对不熟悉使用特定工具链的人友好
2. 环境变量统一通过.env配置，`docker run` 单行命令过多变量可能会敲错

cyfyifanchen:
supported.

PR #82, opened 2024-07-10
Add docker-compose support

Lensual:
#81

### DIFF
--- a/playground/Dockerfile
+++ b/playground/Dockerfile
@@ -1,0 +1,70 @@
+FROM node:18-alpine AS base
+
+# Install dependencies only when needed
+FROM base AS deps
+# Check https://github.com/nodejs/docker-node/tree/b4117f9333da4138b03a546ec926ef50a31506c3#nodealpine to understand why libc6-compat might be needed.
+RUN apk add --no-cache libc6-compat
+WORKDIR /app
+
+# Install dependencies based on the preferred package manager
+COPY package.json yarn.lock* package-lock.json* pnpm-lock.yaml* ./
+RUN \
+  if [ -f yarn.lock ]; then yarn --frozen-lockfile; \
+  elif [ -f package-lock.json ]; then npm ci; \
+  elif [ -f pnpm-lock.yaml ]; then corepack enable pnpm && pnpm i --frozen-lockfile; \
+  else echo "Lockfile not found." && exit 1; \
+  fi
+
+
+# Rebuild the source code only when needed
+FROM base AS builder
+WORKDIR /app
+COPY --from=deps /app/node_modules ./node_modules
+COPY . .
+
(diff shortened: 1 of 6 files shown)


Lensual:
目前 playground 依赖的 `NEXT_PUBLIC_REQUEST_URL` 环境变量只在编译时有效，修改后需要重新 `docker compose build`

next框架不太熟悉

cyfyifanchen:
I had a brief discussion with the other engineers yesterday. We believe that most web developers should already be familiar with how `node` and `nvm` work, which is why we’re not using `docker compose build`. 

However, @Lensual, your point is still valid for developers who aren’t as familiar with web development tech stack . I’ll mark the issue with the question label and see what the other engineers have to say about it.

cyfyifanchen:
Something worth mentioning: we are at the AdventureX 2024 Hangzhou Hackathon. It’s clear that many developers aren’t familiar with the challenges of using Node.js and nvm, especially when dealing with AI, where Python and its environment are predominant.

This makes me think that having a docker-compose file is the right approach. It solves a lot of pain when getting the entire framework up and running.

sunshinexcode:
@Lensual 
docker-compose build astra_playground 会报错, 不能构建出镜像

1.playground/.env.example 后续这个会移除
2..env.sample 这个建议不这样用, docker-compose.yml 改成 docker-compose.yml.example 类似, 使用的时候拷贝一个命名为 docker-compose.yml, 并修改环境变量值

playground/Dockerfile
sunshinexcode, line 1: you can use the following code:

```code
FROM node:20-alpine AS base

FROM base AS builder

WORKDIR /app

COPY . .

RUN npm i && \
    npm run build


FROM base AS runner

WORKDIR /app

ENV NODE_ENV production

RUN mkdir .next

COPY --from=builder /app/.next/standalone ./
COPY --from=builder /app/.next/static ./.next/static

EXPOSE 3000

CMD HOSTNAME="0.0.0.0" node server.js
```

cyfyifanchen:
@Lensual We have concluded that having `docker-compose` is the right thing to do for improving DX. @sunshinexcode will raise a PR for the change. Feel free to review it.

sunshinexcode:
PR: https://github.com/rte-design/ASTRA.ai/pull/123

cyfyifanchen:
@Lensual take a look, if #123 is something you intended to do in the first place? Also from your perspective, which is someone who doesn't know the web development very well, do you think having this `docker-compose` is gonna boost the EX overall?

Lensual:
> @Lensual take a look, if #123 is something you intended to do in the first place? Also from your perspective, which is someone who doesn't know the web development very well, do you think having this `docker-compose` is gonna boost the EX overall?

Yes, it looks good.

PR #88, opened 2024-07-11
Readme update

cyfyifanchen:
* Consistency change
* restructure modals and extension part
* Adding 11labs as option in the config part 
* Adding .next to gitignore

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,13 +69,11 @@ This should start an agent server running on port 8080.
 
 You will need to uncheck "Use Rosetta for x86_64/amd64 emulation on apple silicon" option for Docker if you are on Apple Silicon.
 
-
 <div align="center">
 
 ![ASTRA Docker Setting](https://github.com/rte-design/ASTRA.ai/raw/main/images/docker-setting.gif)
 
 </div>
-
 
 <h3>Connect to Your Agent</h3>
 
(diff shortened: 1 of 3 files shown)


plutoless (changes requested):
the concept of interchangeable is not clear

sunshinexcode (approved):
LGTM

cyfyifanchen:
It means it can be changed to the other extension. I think I do have a little paragraph describes it just right above the table. 

Another word choice is switchable, I ended up choose interchangeable, which I think is little bit better(I also asked GPT about it, he thinks the same). I am open to suggestions on any other better word choices.

plutoless:
> It means it can be changed to the other extension. I think I do have a little paragraph describes it just right above the table.
> 
> Another word choice is switchable, I ended up choose interchangeable, which I think is little bit better(I also asked GPT about it, he thinks the same). I am open to suggestions on any other better word choices.

by far we do not yet have docs to describe how to switch extensions (you will need to manipulate manifest.json). the purpose of that section is to help developer understand the default set up, so emphasizing interchangeable is not very useful, and too early for a beginner.

i would suggest we only describe the default setup here. if needed, you can add an additional section abt how to switch extensions. also i would recommend we do this after our graph designer is ready.

cyfyifanchen:
> > It means it can be changed to the other extension. I think I do have a little paragraph describes it just right above the table.
> > Another word choice is switchable, I ended up choose interchangeable, which I think is little bit better(I also asked GPT about it, he thinks the same). I am open to suggestions on any other better word choices.
> 
> by far we do not yet have docs to describe how to switch extensions (you will need to manipulate manifest.json). the purpose of that section is to help developer understand the default set up, so emphasizing interchangeable is not very useful, and too early for a beginner.
> 
> i would suggest we only describe the default setup here. if needed, you can add an additional section abt how to switch extensions. also i would recommend we do this after our graph designer is ready.

Updated.

README.md
```diff
@@ -73,39 +75,80 @@ You will need to uncheck "Use Rosetta for x86_64/amd64 emulation on apple silico
...
+  <td align="center">openai_chatgpt</td>
+    <!-- <td align="center">✅</td> -->
+    <td align="center">LLM</td>
+    <td align="center">Large Language Modal</td>
+  </tr>
+  <tr>
+  <td align="center">elevenlabs_tts</td>
+    <!-- <td align="center">✅</td> -->
+    <td align="center">Text-to-speech</td>
+    <td align="center">Convert text to audio</td>
```
plutoless, line 119: Elevenlabs Text-to-speech extension?

README.md
```diff
@@ -73,39 +75,80 @@ You will need to uncheck "Use Rosetta for x86_64/amd64 emulation on apple silico
...
+  </tr>
+  <tr>
+  <td align="center">azure_tts</td>
+    <!-- <td align="center">✅</td> -->
+    <td align="center">Text-to-speech</td>
+    <td align="center">Convert text to audio</td>
```
plutoless, line 125: same. maybe add vendor info otherwise the description column looks repetive.

plutoless (changes requested):

README.md
cyfyifanchen: That's right.

README.md
cyfyifanchen: updated

cyfyifanchen:
image

plutoless (approved):

PR #108, opened 2024-07-16
Add Manifest Provider

Lensual:
form #78 

----

Add a manifest.json provider class. Its read manifest.json to memory only, no change to origin file.

- LoadManifest: read manifest from dir.
- injectEnvVar: inject Environment Variable to json content.
- GetManifestJson: give a vendor, return manifestJson.

### DIFF
--- a/server/internal/http_server.go
+++ b/server/internal/http_server.go
@@ -27,13 +27,13 @@ import (
 )
 
 type HttpServer struct {
-	config *HttpServerConfig
+	config           *HttpServerConfig
+	manifestProvider *ManifestProvider
 }
 
 type HttpServerConfig struct {
 	AppId                    string
 	AppCertificate           string
-	ManifestJsonFile         string
 	Port                     string
 	TTSVendorChinese         string
 	TTSVendorEnglish         string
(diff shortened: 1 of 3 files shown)


sunshinexcode:
@Lensual 
这块 manifest 后面还需要内部再讨论, 多个编排怎么存放, 是一个 manifest.json 内部包含多个 graph, 还是以现在多编排文件形式, 如果是多编排文件, 还涉及到如何规范定义命名(命名规则).

plutoless:
thanks for the PR. i think the concept of ManifestProvider is brilliant.
as our extensions keep growing, i guess we will need more flexible ways to adjust single model based on a specific use-case template, instead of creating manifest for every possible combination.

below is my proposal, open for discussion,
1. to provide a usecase manifest template. the template will basically follow the default ASTRA manifest example, while make model extensions as a place holder (e.g. <llm_extension>, <tts_extension>)
2. to provide a vendor config json to manage extension property settings, `{llm: {openai: {...openai extension configs}}}`
3. In manifest provider, based on user request, choose the use case template he want to use, and vendor he want to use, and merge them to get a dynamic generated manifest for agent to use

Before we are reading environment variables and request payload to overwrite property values of manifest json, now we should write to vendor config json (e.g. api keys) or generated manifest file (e.g. voice types)

Let me know what you think.

plutoless:
we have some discussion today and decided to reduce manifest template - there are simply too many of them.
we are planning to separate the graph dynamic settings into a different file so that you can define multiple and choose one of them. pls ignore my previous comments and we will provide an updated solution for this.

PR #137, opened 2024-07-30
feat: support vision for openai

plutoless:


agents/addon/extension/openai_chatgpt_python/openai_chatgpt_extension.py
```diff
@@ -73,11 +79,69 @@ def parse_sentence(sentence, content):
     return sentence, remain, found_punc
 
 
+def yuv420_to_rgb(yuv_data, width, height):
```
wangyoucao577, line 82: Use [PIL convert](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.convert) could be simpler to achieve it? Alternatively, `RGBA` can be outputted by the `agora_rtc` extension, so no conversion is also good to go.

agents/addon/extension/openai_chatgpt_python/openai_chatgpt_extension.py
```diff
@@ -370,6 +363,106 @@ def chat_completions_stream_worker(start_time, input_text, memory):
...
+                                self.send_data(rte, "Let me take a look...", True, input_text)
+                            self.chat_completion_with_vision(rte, start_time, input_text, memory)
+                            return
+                elif chat_completion.choices[0].delta.content is not None:
+                    content = chat_completion.choices[0].delta.content
+            else:
+                content = ""
+
+            # memory is only confirmed when tool is confirmed
+            self.append_memory({"role": "user", "content": input_text})
```
wangyoucao577, line 406: duplicated memory append due to inside the loop?

wangyoucao577 (changes requested):

wangyoucao577 (approved):

PR #156, opened 2024-08-03
chore(): default start graph designer server

sunshinexcode:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ghcr.io/rte-design/astra_agents_build:0.3.3 AS builder
+FROM ghcr.io/rte-design/astra_agents_build:latest AS builder
```
plutoless, line 1: don't use latest. use fixed version.

plutoless: and should use 0.3.4 according to @wangyoucao577

plutoless:
image

PR #177, opened 2024-08-09
chore(): modify default log path

sunshinexcode:


cyfyifanchen (commented):
So, we are still planing to output the logo in the root dir for the sake of DX?

sunshinexcode:
The server logs can be viewed in the console, while the worker logs are located in the directory specified by the LOG_PATH environment variable.

plutoless (approved):

PR #188, opened 2024-08-11
Update README.md

cyfyifanchen:


plutoless (changes requested):
should only merge after vision is by default supported, the code is not yet merged, and it's not yet the default graph

cyfyifanchen:
Should this PR be merged later today or early tomorrow?

Issue #171, opened 2024-08-07
[Feature Request]Support Korean and Japanese


plutoless:
**Is your feature request related to a problem? Please describe.**
Currently no Korean or Japanese is supported.

**Describe the solution you'd like**
Need Korean and Japanese

**Describe alternatives you've considered**
No.

**Additional context**
No.

PR #220, opened 2024-08-19
Add Korean version

Sam-SSTU:
plz check the quick links work or not

thanks!


Sam-SSTU:
#171

cyfyifanchen (approved):
Appreciated for the PR. :) We update the readme quite often, if you can keep eyes on the change, and update the KR version accordingly, that would be awesome. :)

PR #228, opened 2024-08-20
optimize the handling of property.json

sunshinexcode:
1.fix the error in gemini_llm_config attribute settings and prompt
2.optimize the handling of property.json
3.remove openai_proxy_url param
4.process graph for Voice Agent with Vision


Dockerfile
```diff
@@ -5,7 +5,6 @@ ARG SESSION_CONTROL_CONF=session_control.conf
 WORKDIR /app
 
 COPY . .
-COPY agents/property.json.example agents/property.json
```
plutoless, line 8: before we support environment variable in property.json, i believe we still need property.json.example?

README.md
```diff
@@ -63,9 +63,6 @@ In the root of the project, create these files from the examples. They will be u
 ```bash
 # Create .env from the example
 cp ./.env.example ./.env
-
-# Create property.json from the example
-cp ./agents/property.json.example ./agents/property.json
```
plutoless, line 68: same, it's not suitable and dangerous to put property.json out of gitignore now

playground/src/common/request.ts
```diff
@@ -35,12 +35,16 @@ export const apiGenAgoraData = async (config: GenAgoraDataConfig) => {
 
 export const apiStartService = async (config: StartRequestConfig): Promise<any> => {
   const url = `${REQUEST_URL}/start`
-  const { language, channel, userId, voiceType, graphName } = config
+  let { language, channel, userId, voiceType, graphName } = config
+
+  if (graphName == "camera.va.openai.azure.en" && language == "zh-CN") {
```
plutoless, line 40: this is too strange.

plutoless (changes requested):
the change needs to be improved.

Dockerfile
sunshinexcode: no need property.json.example, running the server will not modify the contents of property.json.

playground/src/common/request.ts
sunshinexcode: If handled from the server side, it would lead to hard coding. The current design aims for the graph selection to be specified by the frontend.

PR #232, opened 2024-08-21
issues/231

plutoless:
there should be a flag to control if we would like to print the logs to file or STDOUT because we may have multiple works running, we should add a prefix process id for STDOUT logs

### DIFF
--- a/server/internal/worker.go
+++ b/server/internal/worker.go
@@ -161,8 +161,8 @@ func (w *Worker) start(req *StartReq) (err error) {
 	}
 
 	// Update the prefix with the actual PID
-	stdoutPrefixWriter.prefix = fmt.Sprintf("%s", w.ChannelName)
-	stderrPrefixWriter.prefix = fmt.Sprintf("%s", w.ChannelName)
+	stdoutPrefixWriter.prefix = w.ChannelName
+	stderrPrefixWriter.prefix = w.ChannelName
 	w.Pid = pid
 
 	// Monitor the background process in a separate goroutine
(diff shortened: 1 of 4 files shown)


server/internal/worker.go
```diff
@@ -74,36 +78,107 @@ func getHttpServerPort() int32 {
...
 
+	pid := cmd.Process.Pid
+
 	if pid <= 0 {
 		slog.Error("Worker failed to obtain valid PID after 3 attempts", "requestId", req.RequestId, logTag)
 		return fmt.Errorf("failed to obtain valid PID")
 	}
 
+	// Update the prefix with the actual PID
+	stdoutPrefixWriter.prefix = fmt.Sprintf("%d", cmd.Process.Pid)
```
wangyoucao577, line 164: Is it safe to modify the prefix since cmd has started?

plutoless: yes. a prefix writer is just a proxy for stdout writing, it should be ok to change it.

wangyoucao577: But the `Write` will be called in another Goroutine, right? Personally I'll prefer to use the property json file name as prefix here since it's static and meaningful, even though the downside is too long.

plutoless: @wangyoucao577 yes. the property json file is unnecessarily long and meaningless. i can put channel name as prefix? maybe that's easier to understand. the purpose is to distinguish logs from different processes.

wangyoucao577: @plutoless  OK, it looks like a reasonable trade-off, let's move forward with it.

wangyoucao577:
LGTM

PR #236, opened 2024-08-22
Delete .github/ISSUE_TEMPLATE/feature_request.md

cyfyifanchen:
Removing the feature template for now, less things to go through when a user wants to open a feature request.


plutoless (changes requested):
disagree. having a template is definitely useful. if you believe this is too complicated you can simplify the template.

cyfyifanchen:
I agree that you disagree, that's why I said removing it for now, will add it back later.

PR #242, opened 2024-08-22
refactor & optimize property.json

plutoless:


server/internal/worker.go
```diff
@@ -183,9 +186,9 @@ func (w *Worker) start(req *StartReq) (err error) {
 }
 
 func (w *Worker) stop(requestId string, channelName string) (err error) {
-	slog.Info("Worker stop start", "channelName", channelName, "requestId", requestId, logTag)
+	slog.Info("Worker stop start", "channelName", channelName, "requestId", requestId, "pid", w.Pid, logTag)
 
-	err = syscall.Kill(w.Pid, syscall.SIGTERM)
+	err = syscall.Kill(-w.Pid, syscall.SIGTERM)
```
wangyoucao577, line 191: why negative?

server/main.go
```diff
@@ -54,15 +51,6 @@ func main() {
 		os.Exit(1)
 	}
 
-	// Set graph name map
-	internal.SetGraphNameMap()
```
wangyoucao577, line 58: Same part in `.env` can be removed either?

server/internal/worker.go
plutoless: it means to kill all processes under the process group.
as we are starting the command via "sh -c", it's in fact two processes here, which leaves chances that one is killed while the other is left orphan. so we need to group these two processes and kill them as a group.

server/internal/http_server.go
```diff
@@ -358,6 +358,18 @@ func (s *HttpServer) output(c *gin.Context, code *Code, data any, httpStatus ...
...
+	result := ""
+	for i, elem := range elements {
+		if i > 0 {
+			result += sep
+		}
+		result += elem
+	}
+	return result
+}
+
```
sunshinexcode, line 372: can use strings.Join instead

sunshinexcode:
LGTM

sunshinexcode (approved):
LGTM

wangyoucao577 (approved):

sunshinexcode (approved):
LGTM

PR #247, opened 2024-08-23
1. pre-import llama-index package

plutoless:
2. ensure all processes are killed when SIG_TERM is received

### DIFF
--- a/agents/scripts/install_deps_and_build.sh
+++ b/agents/scripts/install_deps_and_build.sh
@@ -58,6 +58,7 @@ install_python_requirements() {
     done
   fi
 
+  # pre-import llama-index as it cloud download additional resources during the first import
   echo "pre-import python modules..."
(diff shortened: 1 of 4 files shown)


agents/scripts/install_deps_and_build.sh
```diff
@@ -57,6 +57,9 @@ install_python_requirements() {
       fi
     done
   fi
+
+  echo "pre-import python modules..."
+  python3.10 -c "import llama_index.core;"
```
wangyoucao577, line 62: @plutoless It's a reasonable workaround to avoid the unexpected downloading during run. I'm ok with it at the moment, add some comment to describe the issue might be better. However, for long term, may be we can have a pre/post install scripts section in each extension's manifest.json, which may be triggered by tman install? @Halajohn

wangyoucao577 (approved):

PR #248, opened 2024-08-24
feat: use message_collector to replace chat_transcriber

plutoless:
- add a new message_collector
- replace chat_transcriber and take out cmd_conversion
- UI adapt


agents/property.json
```diff
@@ -65,8 +65,8 @@
                     {
                         "type": "extension",
                         "extension_group": "transcriber",
```
wangyoucao577, line 67: extension_group can be renamed either

agents/ten_packages/extension/message_collector/src/extension.py
```diff
@@ -0,0 +1,150 @@
...
+            )
+
+        logger.debug(
+            f"on_data {TEXT_DATA_TEXT_FIELD}: {text} {TEXT_DATA_FINAL_FIELD}: {final} {TEXT_DATA_STREAM_ID_FIELD}: {stream_id} {TEXT_DATA_END_OF_SEGMENT_FIELD}: {end_of_segment}"
+        )
+
+        # We cache all final text data and append the non-final text data to the cached data
+        # until the end of the segment.
+        if end_of_segment:
+            if stream_id in cached_text_map:
```
wangyoucao577, line 117: The LLM's stream_id will be 0 after the cmd_conversion removed, will it be properly handled by the web side?

wangyoucao577 (commented):
So the old chat_transcriber and chat_transcriber_python will be useless and can be removed directly?

agents/ten_packages/extension/message_collector/src/extension.py
plutoless: i searched 999 but it is not used in the entire project. so should be fine

wangyoucao577 (approved):

PR #263, opened 2024-08-30
Feat/api refactor

plutoless:


server/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+
+### POST /start
+This api starts an agent with given graph and override properties. The started agent will join into the specified channel, and subscribe to the uid which your browser/device's rtc use to join.
+
+| Param    | Description |
+| -------- | ------- |
+| request_id  | any uuid for tracing purpose    |
+| channel_name | channel name, it needs to be the same with the one your browser/device joins, agent needs to stay with your browser/device in the same channel to communicate  |
+| chat_uid    | the uid which your browser/device's rtc use to join, agent needs to know your rtc uid to subscribe your audio    |
```
wangyoucao577, line 18: Is `chat_uid` a good name here? maybe `uid` is good enough?

wangyoucao577 (commented):
What's the main purpose for this refactoring? Any problem to solve?

plutoless:
> What's the main purpose for this refactoring? Any problem to solve?

1. move graph logic to front end service, instead of doing it in front-ui on browser
2. provide core agent mgmt apis and docs, so that api can be used without ui
3. add inifinity timeout option when starting worker, so it's easier for developers who starts agent w/o ui (he does not need to keep pinging)

server/README.md
plutoless: uid is bit ambiguous as it can refer to bot uid.. or maybe we should expose both so that it's clearer?

plutoless: few options,
1. human uid
2. adding optional bot_uid, so it's clearer when using

wangyoucao577: Agree that expose both `uid` and `bot_uid`(or maybe `agent_uid`?) seems better.

plutoless: end up with user_uid and bot_uid

wangyoucao577 (approved):

Issue #297, opened 2024-09-29
[DX] Web UI long stuck on first load


seven1240:
**Is your feature request related to a problem? Please describe.**

Since the server is a little far from us, sometimes the first page takes a few seconds to load. The page actually shows very fast and looks like complete, but it still waiting on other resources to load on the background. During the load time, the page is irresponsible, click "Join" has no effect, and looks like "stuck".

**Describe the solution you'd like**

We did a trick to show "Loading ... " instead of "Join" before the page is completely loaded. We can make a pr to review if you guys like it.

https://avatar.ccall.ai/

**Additional context**

The above link is unprotected right now for easy access, please don't abuse it.

plutoless:
fixed.

PR #299, opened 2024-09-29
feat: improve first time loading experience, only enable Join button until it's fully functional

tangpostkarte:
When the login page is loaded, clicking join will not respond and the loading text will be displayed #297


cyfyifanchen:
@tangpostkarte Thanks for the PR. Adding the following state, isn't that right?

image
image

cyfyifanchen (approved):

PR #303, opened 2024-10-01
feat: add a script to generate a nice graph

seven1240:
A simple script to generate a nice graph, currently it only generate the first one, but it could be improved easily.

graph


plutoless (approved):

plutoless:
thank you for the PR!

PR #313, opened 2024-10-09
fix: greeting on user joined

wangyoucao577:


tomasliu-agora (commented):
I think it will be better to have ref count when user join or leave.

agents/property.json
```diff
@@ -680,6 +680,17 @@
                                     }
                                 ]
                             }
+                        ],
+                        "cmd": [
+                            {
+                                "name": "on_user_joined",
```
tomasliu-agora, line 686: I think it will be better to have ref count when user join or leave.

wangyoucao577: added

plutoless (approved):

PR #338, opened 2024-10-16
Issues/214

plutoless:


wangyoucao577:
But how to know the output language?

plutoless:
> But how to know the output language?

no way. i think we can set the language based on browser language though, which should be enough

wangyoucao577 (approved):

wangyoucao577:
> > But how to know the output language?
> 
> no way. i think we can set the language based on browser language though, which should be enough

Sounds good!

PR #340, opened 2024-10-17
256 feature request add channel name in generated file name

plutoless:


wangyoucao577:
It seems that the content is totally different with the pr name?

wangyoucao577 (approved):

Issue #359, opened 2024-10-26
[BUG] Cannot edit extension property field if the value only includes whitespaces


kristol07:
**Describe the bug**
Cannot edit extension property field if the value only includes whitespaces in the playground web app.

**To Reproduce**
1. go to playground web app
2. try to edit any extension property and only input whitespaces
3. when try to re-edit it, cannot be "selected" anymore

![image](https://github.com/user-attachments/assets/de664c0b-8ea1-46d9-b4c5-f25819242252)
![image](https://github.com/user-attachments/assets/373a2078-22ea-4b9d-9ed0-64ac08fdf4dd)

**Expected behavior**
The extension property should be still editable if its last value only include whitespaces

**Screenshots**
See above

**Desktop (please complete the following information):**

 - Browser [chrome]

**Additional context**

kristol07:
> @kristol07 Looks like the fix is merged, would be awesome if you can test and verify. :)

@cyfyifanchen Hi, I made the PR, so yes, it's working now.

cyfyifanchen:
@kristol07 Thanks for your contribution.

PR #360, opened 2024-10-26
fix: whitespace-only properties cannot be editable

kristol07:


kristol07:
related to #359

kristol07:
drop, created new pr

PR #415, opened 2024-11-16
docs: Adding section of windows CRLF to LF line endings

cyfyifanchen:
Adding section of windows CRLF to LF line endings

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ You will need to uncheck "Use Rosetta for x86_64/amd64 emulation on Apple Silico
 ![Docker Setting](https://github.com/TEN-framework/docs/blob/main/assets/gif/docker_setting.gif?raw=true)
 
 #### Windows: Configuring Git to handle line endings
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -171,6 +158,16 @@ Once the build is done, `make run-server` on port `8080`:
 make run-server
 ```
 
+If you are using **Windows**, and see the this error ` /app/agents/bin/start: not found`, you can run the following command to fix it:
+
+```bash
+sed -i 's/\r$//' /app/agents/bin/start
```
plutoless, line 164: you should not be doing this. this will lead to unexpected results.

plutoless (changes requested):
this is not right solution.

- it changes the file, make files affected difficult to track
- the result is unexpected. it may lead to new issues
- it only fix the endings in start script

You should use git auto crlf in this case.

plutoless (approved):

plutoless:
the conflict needs to be resolved @cyfyifanchen

PR #418, opened 2024-11-17
Draft: separate agents

wangyoucao577:
### Typical usage     
     
```bash
# build with default agents
task build

# build with minimax_v2v agents
task build AGENT=examples/minimax_v2v

# clean
task clean

# run server 
task server:run
```    

`make build/clean/run-server` still remains for compatibility, but recommend to use `task build/clean/server:run` instead.


agents/Taskfile.yml
```diff
@@ -0,0 +1,21 @@
+version: '3'
+
+tasks:
+  build:
```
plutoless, line 4: maybe it's better to call it "use agent" which set corresponding folder as main workspace configs?

Taskfile.yml
```diff
@@ -0,0 +1,25 @@
...
+    taskfile: ./agents/Taskfile.yml
+    dir: ./agents
+
+  server:
+    taskfile: ./server/Taskfile.yml
+    dir: ./server
+
+tasks:
+  build:
+    desc: build all
```
plutoless, line 14: is this only useful in cicd?

wangyoucao577: It's equal to `make build`.

agents/Taskfile.yml
wangyoucao577: `task build USE_AGENT=agents/example/minimax_v2v` something like this?

plutoless: no i mean `task use agents/example/minimax_v2v`

Taskfile.yml
plutoless: but if one extension exists in multiple folders, will it be installed for multiple times?

wangyoucao577:
Replaced by https://github.com/TEN-framework/TEN-Agent/pull/433

PR #434, opened 2024-11-23
fix: run go serve load env use Overload

weedge:


wangyoucao577:
What's the benefit of this change?

weedge:
change .env file in project root dir in docker ; if environment variables has defined in sys, can't to load  those in the .env file; just use  environment variables which defined in .env file...........

wangyoucao577:
> change .env file in project root dir in docker container ; if environment variables has defined in sys user mode, can't to load those in the .env file; just use environment variables which defined in .env file.

Does `source .env` work for you?

weedge:
@wangyoucao577 yes, is work,  need set  vars in .env
```
AGORA_APP_ID=***
AGORA_APP_CERTIFICATE=***
```
if not set those env, use user mod defined vars `$AGORA_APP_CERTIFICATE`, 
if vars both in .env and defined vars; so  want use Overload  use vars in .env; :)

wangyoucao577:
> @wangyoucao577 yes, is work, need set vars in .env
> 
> ```
> AGORA_APP_ID=***
> AGORA_APP_CERTIFICATE=***
> ```
> 
> if not set those env, use user mod defined vars `$AGORA_APP_CERTIFICATE`, if vars both in .env and defined vars; so want use Overload use vars in .env; :)

OK. So use `overload` to force to use `.env` seems a little bit no sense here. `export XXX` should has higher priority than `.env`, which is more intuitive. On the other hand, the [godotenv](https://pkg.go.dev/github.com/joho/godotenv#section-readme) also warns that use the `Overload` if needed with caution. So use `Overload` instead of `Load` here may not a good choice.

weedge:
..... but just change the Secrets(such as api key, app token) in .env file, don't to use in user mode defined vars.  unsafe.

PR #438, opened 2024-11-25
refact openai v2v

TomasBack2Future:


agents/ten_packages/system/ten_ai_base/interface/ten_ai_base/config.py
```diff
@@ -44,6 +46,7 @@ def _init(obj, ten_env: TenEnv):
                         val = ten_env.get_property_float(field.name)
                         setattr(obj, field.name, val)
                     case _:
-                        pass
+                        val = ten_env.get_property_to_json(field.name)
```
plutoless, line 49: can we have a more specific type like dict?
better put a comment here.

TomasBack2Future: it should support list and dict, but in case the type never match, because the type will be Dict[str, xxxx]

plutoless: it should just pass if type never match.

agents/ten_packages/extension/glue_python_async/extension.py
```diff
@@ -187,7 +198,21 @@ async def on_data_chat_completion(self, ten_env: AsyncTenEnv, **kargs: LLMDataCo
         messages = []
         if self.config.prompt:
             messages.append({"role": "system", "content": self.config.prompt})
-        messages.extend(self.memory.get())
+        
+        history = self.memory.get()
+        while history:
```
plutoless, line 203: while len(history) > 0?

agents/ten_packages/extension/glue_python_async/extension.py
```diff
@@ -187,7 +198,21 @@ async def on_data_chat_completion(self, ten_env: AsyncTenEnv, **kargs: LLMDataCo
...
+            if history[0].get("role") == "tool":
+                history = history[1:]
+                continue
+            if history[0].get("role") == "assistant" and history[0].get("tool_calls"):
+                history = history[1:]
+                continue
+            
+            # Skip the first tool role
```
plutoless, line 211: why

agents/ten_packages/extension/glue_python_async/extension.py
```diff
@@ -113,17 +97,29 @@ class GlueConfig(BaseConfig):
...
-    sentence_fragment: str = ""
     ten_env: AsyncTenEnv = None
     loop: asyncio.AbstractEventLoop = None
     stopped: bool = False
     memory: ChatMemory = None
-    total_usage: Usage = Usage()
+    total_usage: LLMUsage = LLMUsage()
     users_count = 0
 
+    completion_times = []
```
plutoless, line 117: for metrics maybe create a utility class?

we can use like 
```
metrics.mark("connect_times")
## after some seconds
elapse = metrics.elapse("connect_times")
```

agents/ten_packages/extension/glue_python_async/extension.py
```diff
@@ -343,48 +385,67 @@ async def _send_text(self, text: str) -> None:
...
-                "Authorization": f"Bearer {self.config.token}",
-                "Content-Type": "application/json"
-            }
-            
-            async with session.post(self.config.api_url, json=payload, headers=headers) as response:
-                if response.status != 200:
-                    r = await response.json()
-                    self.ten_env.log_error(f"Received unexpected status {r} from the server.")
-                    return
+        async with aiohttp.ClientSession() as session:
```
plutoless, line 388: maybe cache it to avoid overhead?

plutoless (changes requested):
leave some comments

plutoless (approved):

PR #460, opened 2024-12-04
feat: better module registry

plutoless:
## Purpose
To further enhance module management, solves following issue:
- all addons should have a label for easy readiness
- different llm/v2v extension support different tool set
- only registered modules should i appear, and its type should follow registry configs instead of addon name

## Tasks
- [x] create a module registry class for module management
- [x] update cfg component to respect registry
- [x] only compatible tools should be allowed to choose for different llm/v2v models

### DIFF
--- a/playground/src/components/Chat/ChatCfgModuleSelect.tsx
+++ b/playground/src/components/Chat/ChatCfgModuleSelect.tsx
@@ -47,16 +47,16 @@ export function RemoteModuleCfgSheet() {
 
     const metadata = React.useMemo(() => {
         const dynamicMetadata: Record<string, { type: string; options: { value: string; label: string }[] }> = {};
-    
+
         if (selectedGraph) {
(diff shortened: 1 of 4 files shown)


playground/src/components/Chat/ChatCfgModuleSelect.tsx
```diff
@@ -268,7 +267,7 @@ const GraphModuleCfgForm = ({
                                         <FormItem>
                                             <FormLabel>
                                                 <div className="flex justify-between items-center justify-center ">
```
shczhen, line 269:
```suggestion
                                                <div className="flex items-center justify-center ">
```

dedup `className`

shczhen (approved):
lgtm

```typescript
    export type ModuleType = "stt" | "llm" | "v2v" | "tts";
    export type ToolModuleType = "tool";
```

May consider `enum` to limit variable range.

PR #472, opened 2024-12-12
Fixed typo in property.json weather api key default examples agents

girishg4t:


cyfyifanchen:
Hi, @girishg4t, good catch, It looks like `|` isn’t the only one in the file. Could you update all of them? Thanks for your contribution.

PR #501, opened 2024-12-20
fix build agent and task comman for image booting

debin-ge:
Executable files and environments are missing from the Dockerfile file, causing it to fail to run.
And according to the readme, the startup method is the task command. So when building the image, remove the api executable to run the

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,15 +24,21 @@ RUN apt-get clean && apt-get update && apt-get install -y --no-install-recommend
     python3-dev \
     jq vim \
     ca-certificates \
+    gcc \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -24,15 +24,21 @@ RUN apt-get clean && apt-get update && apt-get install -y --no-install-recommend
     python3-dev \
     jq vim \
     ca-certificates \
+    gcc \
```
wangyoucao577, line 27: It's an runtime dockerfile rather than for build. Here's the build dockerfile that including all the build tools.  https://github.com/TEN-framework/ten_agent_build

PR #553, opened 2025-01-03
Feat/story telling

plutoless:
## Purpose
This PR is to add a new use case to playground. The use case is about an ai agent to tell picture books for child. It will tell story based on your input, and generate image using openai LLVM.

## Tasks
- [x] Support openai_image_generate_tool
- [x] Support story_teller graph
- [x] Support story_teller for all variants
- [x] Add tool into playground registry
- [x] Add docs

## Screenshots
image


agents/ten_packages/system/ten_ai_base/interface/ten_ai_base/const.py
```diff
@@ -3,15 +3,20 @@
 CMD_PROPERTY_TOOL = "tool"
 CMD_PROPERTY_RESULT = "tool_result"
 CMD_CHAT_COMPLETION_CALL = "chat_completion_call"
+CMD_GENERATE_IMAGE_CALL = "generate_image_call"
 CMD_IN_FLUSH = "flush"
 CMD_OUT_FLUSH = "flush"
 
 DATA_OUT_NAME = "text_data"
+RAW_DATA_OUT_NAME = "raw_text_data"
```
wangyoucao577, line 11: How about `content_data` that has an extra `type` to indicate whether text or image buffer or something else?

agents/ten_packages/extension/weatherapi_tool_python/extension.py
```diff
@@ -172,16 +172,24 @@ async def run_tool(
         ten_env.log_info(f"run_tool name: {name}, args: {args}")
         if name == CURRENT_TOOL_NAME:
             result = await self._get_current_weather(args)
-            # result = LLMCompletionContentItemText(text="I see something")
-            return {"content": json.dumps(result)}
+            return LLMToolResultNormal(
+                type="normal",
```
wangyoucao577, line 176: How about `history_only`? Which is a little bit special compare with `requery`.

agents/ten_packages/extension/weatherapi_tool_python/extension.py
```diff
@@ -172,16 +172,24 @@ async def run_tool(
         ten_env.log_info(f"run_tool name: {name}, args: {args}")
         if name == CURRENT_TOOL_NAME:
             result = await self._get_current_weather(args)
-            # result = LLMCompletionContentItemText(text="I see something")
-            return {"content": json.dumps(result)}
+            return LLMToolResultNormal(
```
wangyoucao577, line 175: `LLMToolResult` may be good enough, while it's better to have some introduction about what `requery` expects LLM does.

agents/examples/demo/property.json
```diff
@@ -1252,6 +1252,265 @@
...
+          {
+            "type": "extension",
+            "name": "message_collector",
+            "addon": "message_collector",
+            "extension_group": "transcriber",
+            "property": {}
+          },
+          {
+            "type": "extension",
+            "name": "message_collector2",
```
wangyoucao577, line 1326: Use same `message_collector` to processes different data will be better.

wangyoucao577 (approved):

PR #559, opened 2025-01-08
feat: add AWS Nova multimodal llm for support realtime

zhuermu:
Add a addon for support Voice Agent Realtime such as: Gemini2.0. User input voice frame and  video frame and output voice and text.
The aws Nova multimodal support file，image and video as input output text. So we base on the extension of the Ten framework:  agora_rtc+STT+Nova+TTS and we can achieve real-time like effect.
Here are the test results:
image


plutoless:
maybe it's worth to change https://github.com/TEN-framework/TEN-Agent/blob/main/playground/src/common/moduleConfig.ts
to register nova as one of v2v modules. so that developers can easily switch and try it out in playground?

plutoless:
i think the bedrock_v2v_python is still consuming input text and produce output text. what is different from the old bedrock_llm_python extension? or should we just update the existing extension to use your code?

plutoless (changes requested):
thanks for the pr! but i'm bit confused as the graph tells bedrock_v2v_python is still a text-in text-out llm extension. if needed i would suggest to make your bedrock_v2v_python a updated version of bedrock_llm_python. does this make sense?

zhuermu:
> thanks for the pr! but i'm bit confused as the graph tells bedrock_v2v_python is still a text-in text-out llm extension. if needed i would suggest to make your bedrock_v2v_python a updated version of bedrock_llm_python. does this make sense?

The new bedrock Nova is multimodal. I agree your idea, I try to update the bedrock_llm_python. Thanks .

PR #589, opened 2025-02-13
fix: Add a graph named "voice_assistant_multimodal" in the property.son of default.

zhuermu:
Some developers are confused about debugging the Amazon Nova multimodal model, so we are adding a copy of the demo graph to the default configuration.


cyfyifanchen:
LGTM

cyfyifanchen (approved):
LGTM

plutoless (changes requested):
you should register nova as a llm module in playground moduleConfig.ts
the concept of playground is using the same connections while developers are able to switch modules. if you believe video feature is needed, it should rather be a option feature flag for llm module.
to conclude,
1. add nova to llm registry in moduleConfig.ts
2. in moduleConfig.ts for llm module add an option `modalities=["image", "text"]`
3. change ChatCfgModuleSelect.tsx so that a connection from rtc to llm will be added when a llm module accepts image input

plutoless (approved):

PR #638, opened 2025-03-26
translated readme

seanwhen:
Updated 6 languages md according to README.md


cyfyifanchen:
@seanwhen let's do them one file at the time, it's easier for us to review, this said, let's focus on `README-CN.md` first.

PR #657, opened 2025-04-02
README-CN Agora API Registration

seanwhen:
National user api registration


cyfyifanchen:
@seanwhen Thanks for the PR, steady and fast.  A little advice to make it even more outstanding is that having different a branch name each time when you make a PR, that way, the history of each commit doesn't stack together. Also, having a distinguished name for each PR per different features is generally considered the best practice.

cyfyifanchen (approved):

PR #903, opened 2025-05-29
fix: fixed the incompatibility issue with the new version of ElevenLab…

Faustrazor:
In the TTS step, the AsyncElevenLabs.generate function was affected by a breaking change in the SDK. According to the official documentation, the latest SDK version (v1.58.0) was released to fix this issue. I updated the streaming TTS call implementation based on the new API reference.


plutoless (approved):

plutoless:
LGTM. thank you for the PR!

PR #1128, opened 2025-07-24
fix: adding script bin/start to all the graphs

cyfyifanchen:


plutoless:
this is no longer needed

PR #1150, opened 2025-07-28
feat: cloudsway tts extension

pikerbright:
Add cloudsway tts extension.


plutoless (approved):

plutoless (changes requested):
changing network mode to host will lead to port conflict for demo and playground

PR #1200, opened 2025-08-05
feat: add aws_asr_python extension

ygcaicn:
iShot_2025-08-05_16 23 03


ai_agents/agents/ten_packages/extension/aws_asr_python/tests/configs/property_en.json
```diff
@@ -0,0 +1,12 @@
+{
+    "dump": false,
+    "finalize_mode": "disconnect",
+    "params": {
+        "region": "us-west-2",
```
sunxilin: region should also be read from env var

ai_agents/agents/ten_packages/extension/aws_asr_python/tests/configs/property_en.json
```diff
@@ -0,0 +1,12 @@
+{
+    "dump": false,
```
sunxilin: it's better that the 'dump' flag not set by default

PR #1218, opened 2025-08-08
feat: adding gpt-5

cyfyifanchen:


plutoless (changes requested):
can we reuse exisiting openai extension?

cyfyifanchen:
> can we reuse exisiting openai extension?

Tested, didn’t work directly. The max_tokens and both pentitles parameters need to be disabled, otherwise it will throw errors. Considering that 4o is being called in many places within the project, it’s better to keep both 4o and gpt-5 in the project for now. As gpt-5 becomes more stable, we can gradually replace 4o with gpt-5.

cyfyifanchen:
Submitted another PR.

PR #1352, opened 2025-08-26
add-cn-readme

AprilJW:
add chinese readme


cyfyifanchen:
@AprilJW thanks for the PR, a couple of of things.

1. Let’s use English instead of Chinese, Github is an international community, we've got other folks don't speak Chinese here. 
2. Let’s follow the conventional commit specification.
3. Your PR should synchronize the latest English version with the Chinese version.

cyfyifanchen:
#1344

cyfyifanchen (changes requested):
Good fix, a couple more things let's pay attention to

1. We aren't creating new README-CN, instead. We want to sync the dynamic content from README to README-CN
2. There are parts within README-CN they are unnecessary in README.

:)

cyfyifanchen:
Very good following up work and fix, pretty much all the things needed are there, one thing need a little more work tho:

image

cyfyifanchen (approved):

halajohn (approved):

PR #1493, opened 2025-09-13
fix: tencent asr signature

ygcaicn:


sunxilin:
cherrypick to another branch

PR #1505, opened 2025-09-16
fix: fix interrupt issue for tencent tts extension

seymourtang:
* fix: fix the issue that cannot interrupt tts
* chore: adjust the calculation method of ttfb and request_event_interval
* feat: update TTS extensions to include current_turn_id in audio start calls


seymourtang:
@YiminW

YiminW (approved):

plutoless (approved):

PR #1809, opened 2025-12-01
feat: add a python test wrapper supporting AAA-style tests

anexpn:
- [ ] Add tests examples for test_wrapper

### DIFF
--- a/core/src/ten_runtime/binding/python/interface/ten_runtime/test_wrapper.py
+++ b/core/src/ten_runtime/binding/python/interface/ten_runtime/test_wrapper.py
@@ -1,0 +1,600 @@
+#
+# Copyright © 2025 Agora
+# This file is part of TEN Framework, an open source project.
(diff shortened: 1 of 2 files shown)


core/src/ten_runtime/binding/python/interface/ten_runtime/test_wrapper.py
```diff
@@ -0,0 +1,600 @@
...
+# Copyright © 2025 Agora
+# This file is part of TEN Framework, an open source project.
+# Licensed under the Apache License, Version 2.0, with certain conditions.
+# Refer to the "LICENSE" file in the root directory for more information.
+#
+import asyncio
+import traceback
+from typing import Callable, Any, AsyncGenerator
+
+from ten_runtime import (
```
sunxilin: There might be circular dependencies; changing to relative path dependencies should resolve this.

anexpn: done

PR #1892, opened 2025-12-17
feat: add doodler a doodle board for kids

cyfyifanchen:


plutoless:
why not just use openai_image_generate_tool?

cyfyifanchen:
> why not just use openai_image_generate_tool?

cuz gpt image 1.5 doesn't work with `openai_image_generate_tool`, the tool only works with gpt image 1.

plutoless (approved):

Issue #1907, opened 2025-12-22
[BUG]  .model: unsupported conversion from `null` to `string`


Huangshan16:
### Description

task run example _voice-assistant-video_  warning "extension.c:1017[0m [37m[llm] See in cmd::chat_completion with invalid schema: .model: unsupported conversion from `null` to `string`"

then LLM can't reply

<img width="1812" height="853" alt="Image" src="https://github.com/user-attachments/assets/cf76e26c-b3c3-4d7f-9a4b-275dff53b6b4" />

### Environment

win11, WSL2, image version: 0.7.12

### Steps to reproduce

cd ai_agents
cp ./.env.example ./.env
docker compose up -d
docker exec -it ten_agent_dev bash
cd agents/examples/voice-assistant-video
task install
task run

### Expected behavior

LLM can reply

### Severity

Critical

PR #1909, opened 2025-12-22
fix: .model: unsupported conversion from null to string (#1907)

Huangshan16:


plutoless:
the model field has been removed in previous ten_ai_base version. pls try update the deps and run again

PR #1946, opened 2026-01-05
fix: azure asr fatal error

sunxilin:


zhangjie02agoraio (approved):

ai_agents/agents/integration_tests/asr_guarder/tests/test_reconnection.py
```diff
@@ -283,13 +287,34 @@ async def on_data(self, ten_env: AsyncTenEnvTester, data: Data) -> None:
             if not self._validate_error_format(ten_env, json_data):
                 return
 
+            # Check if this is a fatal error (won't retry)
+            error_code: int = json_data.get("code", 0)
+            self.error_codes.append(error_code)
+
+            # Fatal errors are defined by ModuleErrorCode.FATAL_ERROR = -1000
+            # Non-fatal errors are defined by ModuleErrorCode.NON_FATAL_ERROR = 1000
+            is_fatal = error_code == -1000
```
linyudoa: Should we use ModuleErrorCode.FATAL_ERROR.value, ModuleErrorCode.NON_FATAL_ERROR.value rather than a magic number?

ai_agents/agents/integration_tests/asr_guarder/tests/test_reconnection.py
```diff
@@ -341,3 +366,39 @@ def test_reconnection(extension_name: str, config_dir: str) -> None:
     assert (
         error is None
     ), f"Test failed: {error.error_message() if error else 'Unknown error'}"
+
+    # Verify error handling behavior based on error sequence
+    if tester.fatal_error_received:
+        # FATAL error: once fatal error occurs, should not retry anymore
+        # Check that fatal error (-1000) is present in error codes
```
linyudoa: Same as above if the suggestion applies

ai_agents/agents/integration_tests/asr_guarder/tests/test_reconnection.py
sunxilin: Good suggestion! FATAL_ERROR enum is defined in `ten_ai_base` module, which is also a TEN system package each ASR Extension depends on. The `asr_guarder` always indirectly depends on `ten_ai_base`, so importing the corresponding Python module is reasonable.

wangyoucao577 (approved):

plutoless (approved):

PR #2003, opened 2026-01-20
fix: log emitted to otlp

sunxilin:


core/src/ten_rust/src/log/otel.rs
```diff
@@ -317,6 +374,24 @@ pub fn create_otlp_layer(
...
+    // level based on user config.
+    let recursion_filter = Targets::new()
+        .with_target("opentelemetry", LevelFilter::OFF)
+        .with_target("tonic", LevelFilter::OFF)
+        .with_target("h2", LevelFilter::OFF)
+        .with_target("hyper", LevelFilter::OFF)
+        .with_target("tower", LevelFilter::OFF)
+        .with_target("reqwest", LevelFilter::OFF)
+        .with_target("rustls", LevelFilter::OFF)
+        .with_default(LevelFilter::TRACE);
```
linyudoa: Wondering whether a whitelist will be better if the rust packages we care about is very limited in number...If more helper packages are introduced and generate exceesive logs, someone have to remember to write some code here to turn it off? Though it doesn't matter too much now...

halajohn (approved):

PR #2038, opened 2026-02-05
Change the openai base_url in voice-assistant's property.json 

hedongya:
Fix the issue of LLM key errors. The openai base_url in the voice-assistant's property.json file is hardcoded as "https://api.openai.com/v1". If the openai parameter in .env is configured with parameters for a Chinese model, it will cause key errors. Now, modify the openai base_url to reference a variable instead.


plutoless:
umm.. i think your code change does not match with your pr description. would you mind check a little bit more on the diff?

PR #2040, opened 2026-02-05
fix(TEN-5338): send sample rate to server

YiminW:
Description: openai tts server doesn't support sample rate, but other tts vendor based in openai protocal support it, so send it always, if vendor doesn't support will ignore it.


wangyimin-agora:
image
image

plutoless (approved):

PR #2047, opened 2026-02-11
feat: add voice-assistant-with-EverMemOS example with memory integration

AI-J-IN:
- Add new voice assistant example with EverMemOS integration


plutoless (approved):
LGTM. thanks for the contribution!

PR #2055, opened 2026-02-17
fix(anam_avatar_python): handle vercel bypass token via configuration

RinZ27:
## Summary
Hardcoded Vercel bypass tokens in the Anam extension were caught during a recent audit. Leaving these visible in the source code is a bit risky since they're meant to bypass deployment protections, so I decided to shift them over to the extension configuration.

Moving the token into `AnamAvatarConfig` allows users to set it through property settings or environment variables now. Updated the manifest and default property files as well to make sure everything stays consistent.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Testing
- [x] Manual testing completed
- [ ] All tests pass (CI failure in `claude-review` appears to be a permission issue with the workflow itself, unrelated to these changes)

## Documentation
- [x] Documentation updated (manifest.json and property.json)
- [x] Examples provided if needed

## Breaking Changes
None. Default behavior is preserved, but users now have the option to configure the bypass token externally.


plutoless:
LGTM thank you for your pr!

plutoless (approved):

PR #2188, opened 2026-06-13
fix(assemblyai_asr): route unknown messages to on_event not on_error

plutoless:
## Problem

The AssemblyAI ASR extension crashed on normal streaming traffic with:

```
E _handle_message@recognition.py:130 [asr] [AssemblyAI] Error processing message: 1 validation error for ModuleError
message
  Input should be a valid string [type=string_type, input_value={'type': 'SpeechStarted',..., 'confidence': 0.38146}, input_type=dict]
```

## Root cause

AssemblyAI's v3 streaming API sends informational messages such as `SpeechStarted`. In `recognition.py:_handle_message` only `Begin`/`Turn`/`Termination` are handled; everything else fell into the `else` branch which called `self.callback.on_error(message_data)` — passing the **raw dict**. In `extension.py:on_error` that dict was fed into `ModuleError(message=...)`, whose `message` field is typed `str`, so Pydantic raised the `string_type` validation error (caught and logged at `recognition.py:130`).

Two defects:
- **Semantic** — `SpeechStarted` is not an error; routing it to the error path would also emit a spurious ASR error downstream.
- **Type** — a dict was passed where `on_error` contractually requires a `str`.

## Fix

- `recognition.py` — unrecognized messages are now routed to `on_event` (accepts a dict, logs only) instead of `on_error`.
- `extension.py` — fixed a missing `f` prefix on the `on_event` log line (now on the active path) so the payload interpolates.
- Added `tests/test_handle_message.py` asserting `SpeechStarted` and other unknown message types route to `on_event`, not `on_error`.

## Verification

The local environment lacks `pytest` and the runtime deps (`websockets`, `ten_runtime`, `ten_ai_base`), so the project harness could not run here. The fix was verified by loading the real `recognition.py` with those deps stubbed and exercising `_handle_message`:

- Before: `on_error` called with a `dict` (reproduces the crash trigger).
- After: `on_error` not called; `on_event` called with the `SpeechStarted` dict.

The added unit tests run in CI where the deps are available.


diyuyi-agora:
The lint needs to be fixed first

diyuyi-agora:
tests/test_handle_message.py::test_connection_closed_passes_string_args_to_on_error  failed.

ai_agents/agents/ten_packages/extension/assemblyai_asr_python/tests/test_handle_message.py
```diff
@@ -0,0 +1,136 @@
...
+    on_error feeds ModuleError(message=str)/ModuleErrorVendorInfo(code=str), so
+    passing the raw exception/int code would raise a Pydantic validation error.
+    """
+    from websockets.exceptions import ConnectionClosed
+
+    callback = _RecordingCallback()
+    recognition = _make_recognition(callback)
+    recognition.websocket = _RaisingWebSocket(
+        ConnectionClosed("connection closed: 4001 (some reason)")
+    )
```
diyuyi-agora, line 131: With websockets~=14.0  ConnectionClosed does not accept a string. Its constructor requires:

ConnectionClosed(rcvd: Close | None, sent: Close | None, ...)

Error: 
FAILED tests/test_handle_message.py::test_connection_closed_passes_string_args_to_on_error - TypeError: ConnectionClosed.__init__() missing 1 required positional argument: 'sent'

diyuyi-agora (approved):

PR #2193, opened 2026-06-24
feat(tts): add gradium tts support

BenWeekes:
## Summary

Adds the **Gradium TTS** extension (`gradium_tts_python`), a websocket streaming TTS built on `AsyncTTS2BaseExtension`, plus a `voice_assistant_gradium` example graph.

- **Vendor client** (`gradium_tts.py`): websocket setup → `ready` → `text`/`end_of_stream` → `audio`/`error`, with auth-error classification (401/403/1008 → fatal).
- **Extension** (`extension.py`): streams audio through a single finalize path (`_finalize_request` → `finish_request`); request-level TTFB metric (emitted once per request even across multiple vendor segments); audio framing prefers the server-`ready` sample rate, falling back to config.
- **Config** (`config.py`): pydantic model with param normalization, output-format/sample-rate reconciliation, and api-key redaction in logs.
- **Example wiring**: new `voice_assistant_gradium` predefined graph (deepgram ASR + openai LLM + gradium TTS), manifest dependency, regenerated lockfile.

## Tests

Standalone suite mirrors the sibling TTS extensions (basic audio, flush, state machine, robustness, params, metrics, error handling), plus gradium-specific coverage for its text-batching and per-request socket model.

- `task test-extension EXTENSION=agents/ten_packages/extension/gradium_tts_python` → 27 passed
- `task tts-guarder-test EXTENSION=gradium_tts_python CONFIG_DIR=tests/configs` → 15 passed, 1 skipped

## Follow-ups (non-blocking)

- Add punctuation-only / list-transition robustness regressions to match the documented robustness matrix.
- Add an explicit 48 kHz config/test if we advertise `pcm_48000`.
- If a future refactor reuses one websocket across segments, revisit reconnect/backoff coverage.


ai_agents/agents/ten_packages/extension/gradium_tts_python/config.py
```diff
@@ -3,86 +3,128 @@
...
         if "dump" in params:
-            self.dump = params.pop("dump")
+            self.dump = bool(params.pop("dump"))
+
         if "dump_path" in params:
-            self.dump_path = params.pop("dump_path")
+            self.dump_path = str(params.pop("dump_path"))
 
-        self.params = params
+        self.output_format = self._normalize_output_format(
```
YiminW: please check these two functions.
which foramt and sample rate will be used in set both?
And we only support pcm format

ai_agents/agents/ten_packages/extension/gradium_tts_python/config.py
```diff
@@ -3,86 +3,128 @@
...
     def websocket_url(self) -> str:
-        """Build the websocket URL based on region or explicit base_url."""
-        if self.base_url:
-            return self.base_url
-
-        region = (self.region or "us").lower()
-        if region not in {"us", "eu"}:
-            region = "us"
-        return f"wss://{region}.api.gradium.ai/api/speech/tts"
+        return self.base_url.strip() or GRADIUM_DEFAULT_WS_URL
```
YiminW: base url means the basic url, and we should add path to it in code.
I think it should be url in this config

YiminW (changes requested):

BenWeekes:
### TTS guarder + standalone test records (commit 931cc2449)

Addressed both review comments on `config.py`:
- **sample_rate is the single source of truth** — Gradium only supports PCM, so the wire `output_format` is now derived as `pcm_<sample_rate>` and is no longer a user-settable knob (removed the two conflicting normalize/derive helpers). This matches the closest peer `minimax_tts_websocket` and the convention across other TTS extensions (`sample_rate` is the canonical rate knob).
- **`base_url` → `url`** — the field holds the full websocket endpoint and is used as-is (no path appended in code), matching the `url` convention used by the modern WS extensions (`minimax`, `vibevoice`).

#### TTS guarder — `task tts-guarder-test EXTENSION=gradium_tts_python`
Run against the **real Gradium API**, from the branch worktree (container `/app` is a bind mount of this branch). **15 passed, 1 skipped in 363.94s.**

| Case | Result |
|------|--------|
| test_append_input | ✅ passed |
| test_append_input_stress | ✅ passed |
| test_append_input_without_text_input_end | ✅ passed |
| test_append_interrupt | ✅ passed |
| test_basic_audio_setting | ✅ passed |
| test_corner_input | ✅ passed |
| test_dump | ✅ passed |
| test_dump_each_request_id | ✅ passed |
| test_empty_text_request | ✅ passed |
| test_flush | ✅ passed |
| test_interleaved_requests | ✅ passed |
| test_invalid_required_params | ✅ passed |
| test_invalid_text_handling | ✅ passed |
| test_metrics | ✅ passed |
| test_miss_required_params | ✅ passed |
| test_subtitle_alignment | ⏭️ skipped (disabled by default; optional — only for providers that emit word/segment timing, which Gradium does not) |

#### Standalone suite — `task test-extension EXTENSION=...gradium_tts_python`
**28 passed** (includes a new `test_output_format_derived_from_sample_rate_only` that locks in: a user-supplied `output_format` is ignored, `sample_rate` drives the derived `pcm_<rate>`, and it is not leaked as a vendor passthrough param).

Both runs were executed on the branch code; black format-check is clean.

ai_agents/agents/ten_packages/extension/gradium_tts_python/config.py
BenWeekes: Fixed in 931cc2449. Since Gradium only supports PCM, sample_rate is now the single source of truth: I removed output_format as a settable knob and derive it as pcm_<sample_rate> on the wire, so format and sample rate can never disagree. The two helper functions you flagged (_normalize_output_format / _sample_rate_from_output_format) are gone, and a user-supplied output_format is ignored (not double-sent as a vendor param). This matches the closest peer minimax_tts_websocket (PCM-only) and the sample_rate convention used across the other TTS extensions. Added test_output_format_derived_from_sample_rate_only to lock it in.

ai_agents/agents/ten_packages/extension/gradium_tts_python/config.py
BenWeekes: Fixed in 931cc2449. Renamed the config field base_url -> url (and updated manifest, property.json, test configs, README, tests). The endpoint is used as-is via websocket_url() with no path appended in code, which matches the url convention used by the modern WS extensions (minimax, vibevoice).

BenWeekes:
### Update — commits `3bb3e909` and `b2c1b819`

#### On the `on_data` override (re: "the base class already implements `on_data`, so it shouldn't be needed")

I looked into removing the override and it turns out it's **load-bearing** — removing it drops text on interleaved requests. Keeping it, with a comment now explaining why.

**Mechanism.** Gradium is the only TTS extension that *batches* multiple same-`request_id` segments into a single vendor request (it accumulates `pending_text` and flushes on a sentence boundary or `text_input_end`, then deletes the per-request context on finalize). The base class buffers messages for a not-yet-active `request_id` and, in `finish_request`, re-queues them to the **tail** of its input queue. For interleaved requests this can hand `request_tts()` a request's `text_input_end` segment **before** an earlier segment of the same request — which flushes and deletes the context early, dropping the earlier text.

`ingress_messages` is a per-`request_id` FIFO captured at `on_data` arrival time and replayed in order in `request_tts()`, which neutralises that reordering.

**Evidence.** I empirically removed the override and ran the standalone suite — `test_interleaved_requests_keep_separate_buffers` fails, dropping `interleave_req_2`'s first segment:
```
At index 1: ('interleave_req_2', 'It also finishes in one vendor request.')
  != ('interleave_req_2', 'Request two starts now and stays separate. It also finishes...')
```

**Why the other extensions don't need this:** I checked `minimax`, `cartesia`, `elevenlabs`, and `xai_tts` — they all send each segment to the vendor *immediately* inside `request_tts`, with no per-request text buffer, so an out-of-order `text_input_end` can't drop earlier text. Gradium's per-request-socket model makes tiny per-fragment requests expensive, which is why it batches — and why it alone is exposed.

`b2c1b819` adds a thorough comment on `on_data`/`ingress_messages` documenting this invariant and pointing at the covering test.

**Note:** the cleaner long-term fix is in the base class — if `finish_request` preserved per-request arrival order when re-releasing buffered messages (rather than re-queuing to the tail), gradium could drop `ingress_messages` entirely and every extension would be safe. Happy to file that as a separate issue against `ten_ai_base`.

#### websockets 14 close-code (`3bb3e909`)

Separately fixed a latent bug: `websockets~=14.0` exposes the close code on `exc.rcvd`/`exc.sent` (each a `Close` frame with `.code`), not a top-level `exc.code`. The clean-close path in `_iter_messages` relied on `getattr(exc, "code", ...)`, so a real `1000` end-of-stream close would be misclassified as `EVENT_TTS_ERROR` instead of `EVENT_TTS_END` (the existing test passed only because it hand-built an exception with a top-level `.code`). Added a `_close_code` helper that reads `rcvd`/`sent` first and falls back to `.code`, plus a regression test using the real websockets-14 exception shape.

Standalone suite: **29 passed** (28 + the new regression test); `black` clean.

YiminW:
I don't think Gradium should batch all text segments with the same `request_id` before sending them to the vendor.

A few reasons:

1. This is not how the other TTS extensions behave. Cartesia / ElevenLabs / xAI all forward each incoming `TTSTextInput` segment to the vendor/client layer as it arrives, and use `text_input_end` only to finalize the request. Gradium is currently different: it accumulates `pending_text` and only sends when the buffer reaches a sentence/size threshold or `text_input_end`.

2. I don't think the `AsyncTTS2BaseExtension` ordering issue described here is accurate. `tts2.py` buffers messages from *other* `request_id`s while one request is being processed, and `finish_request()` releases the next buffered request back to the input queue in FIFO order. It serializes interleaved requests, but it should not reorder messages within the same `request_id` such that a `text_input_end=True` segment is processed before earlier segments of that same request. If we have a concrete repro for that, it should probably be fixed/tested in `AsyncTTS2BaseExtension`, not worked around by overriding `on_data()` in the Gradium extension.

3. Gradium's official streaming guidance does not recommend waiting for the whole logical request before sending text. Their LLM-to-TTS recipe says to open one TTS WebSocket before the LLM starts, buffer tokens into word/phrase chunks, send each chunk as a `text` message, add `<flush>` at sentence/end boundaries, and read audio concurrently:
https://docs.gradium.ai/guides/recipes/llm-to-tts.md

   The documented chunking rules are:
   - send complete words, phrases, or sentences
   - keep punctuation attached to the preceding word
   - avoid one-token-per-message unless tokens are already clean word chunks
   - add `<flush>` at natural boundaries
   - avoid frequent flushes because they can make prosody choppy

   In our pipeline, the LLM already guarantees that every emitted segment ends with punctuation, so the incoming segments should already satisfy Gradium's recommended chunking constraints.

BenWeekes:
### Update — stream segments immediately (commit `89da4f70`)

This reworks the extension to **forward each `tts_text_input` segment to Gradium as it arrives**, instead of accumulating segments in a local buffer. It resolves the earlier review feedback about overriding `on_data`, and fixes the dropped-audio behaviour seen in manual testing.

#### What changed
- **`gradium_tts.py`** — the client is now a persistent **per-request streaming session**: `start_session()` (connect + setup + `ready`) → `send_text(segment)` for each segment → `end_input()` (`end_of_stream`) → audio read concurrently via `audio_events()`. (Previously it opened one websocket per `get()` call.)
- **`extension.py`** — `request_tts()` forwards each segment immediately over a single session for the request; a background reader streams audio back and finalizes on stream end. **Removed** the `pending_text` batching, `_should_flush_pending_text`, the `on_data` override, `ingress_messages`, and the per-request `RequestContext` dict.
- Retains the websockets-14 close-code fix (`exc.rcvd`/`exc.sent` vs top-level `exc.code`).

#### Why
- **Matches the other websocket TTS extensions.** Cartesia, ElevenLabs, xAI and Minimax all forward each segment to the vendor immediately and use `text_input_end` only to finalize. Gradium was the only one batching, which is also what forced the non-standard `on_data` override.
- **Matches Gradium's official LLM-to-TTS guidance** (open one socket, send each chunk as a `text` message, read audio concurrently). Our LLM already emits punctuation-terminated segments, so each incoming segment is a valid chunk to send as-is.
- **Fixes dropped audio.** Holding text in a local buffer that a finalize could flush-and-discard was dropping sections of speech. With immediate forwarding there is no buffer to lose, and the base class owns queuing/ordering — so the `on_data` override is no longer needed.

#### Verification
- **Manual:** `voice_assistant_gradium` graph — clean multi-turn conversation; logs show each request completing with `reason=REQUEST_END`, one TTFB metric per request, no errors/warnings, api_key redacted.
- **TTS guarder (real Gradium API):** 15 passed, 1 skipped (incl. `test_interleaved_requests`, `test_flush`, `test_append_interrupt`, `test_corner_input`, `test_metrics`).
- **Standalone suite:** 29 passed (rewritten to the streaming API; adds immediate-forwarding, single-session-per-request, and punctuation-only coverage).
- **Formatting:** `task check` (black, line-length 80) clean.

Net change: roughly −260 lines; the extension is now simpler and consistent with the rest of the TTS extensions.

wangyoucao577 (approved):

PR #2194, opened 2026-06-24
feat: add Mistral Voxtral TTS extension

TiagoAgora:
## Summary

Adds a new TTS extension, `mistral_tts_python`, integrating Mistral's Voxtral text-to-speech via the OpenAI-compatible `/v1/audio/speech` endpoint.

- Built on `AsyncTTS2HttpExtension` (HTTP TTS mode), following the extension development guide.
- Requests a self-describing WAV stream and converts it to PCM16 mono on the fly (handles int16/int24/int32 and IEEE-float payloads); audio output at 24 kHz.
- Forwards vendor params through unchanged (`model`, `voice_id`, `ref_audio`, …); API-key auth via the `Authorization` header.
- Handles cancellation/flush, content-moderation/auth errors, and TTFB metrics.

## Testing

Standalone unit tests pass in the dev container (`tman install --standalone` + `tests/bin/start`): **13/13 passing** — covering config defaults/validation, URL/base_url resolution, headers, audio dump, flush/cancel, invalid-key handling, reconnect robustness, metrics, and the request state machine.

Live API validation against `api.mistral.ai` has not been run yet.


ai_agents/agents/ten_packages/extension/mistral_tts_python/config.py
```diff
@@ -0,0 +1,99 @@
...
+        if "model" not in self.params:
+            self.params["model"] = DEFAULT_MODEL
+
+        # Remove input if present (will be set from text)
+        if "input" in self.params:
+            del self.params["input"]
+
+        # Always request a self-describing WAV stream; the client converts it
+        # to PCM16 mono regardless of the underlying WAV sample format.
+        self.params["response_format"] = DEFAULT_RESPONSE_FORMAT
```
YiminW, line 53: Image
it seems support pcm format?

TiagoAgora: Good catch, switched to `pcm`.

You're right that `pcm` is supported; the reason I'd originally gone with `wav` is that Voxtral's `pcm` is a **headerless float32 LE** stream (24 kHz mono), not the int16 PCM16 that TEN's `pcm_frame` contract expects, so it can't be passed straight through either way. But requesting `pcm` is the better choice: it's lower latency (no container header to buffer before the first audio) and simpler, so the client now requests `pcm` and rescales each float32 sample to int16 (`Float32ToPcm16`), dropping the WAV chunk-parsing path entirely. Non-finite samples map to silence so a corrupt stream can't crash conversion. Tests updated to stream headerless float32 mocks.

Note: the full standalone suite runs in the dev container (`tman install --standalone` + `tests/bin/start`); I verified the float32→pcm16 conversion (round-trip, split-sample buffering, clamping/NaN) in isolation here.

YiminW:
HI Tiago, please run tts-guarder for your new tts extension and add all cases passed sanp here

TiagoAgora:
## TTS guarder results — `mistral_tts_python`

Ran `task tts-guarder-test EXTENSION=mistral_tts_python CONFIG_DIR=tests/configs` against the live Mistral / Voxtral API.

**14 passed · 1 skipped · 1 failed**

mistral_tts_guarder_results


⏭️ **`test_subtitle_alignment`** — skipped; Voxtral exposes no word-level timing.

❌ **`test_interleaved_requests`** — traced to a pre-existing bug in `ten_ai_base` (`tts2.py`), **not this extension**. When `finish_request()` releases buffered interleaved requests it pre-sets `_processing_request_id`, so `_process_input_queue()` skips the `QUEUED → PROCESSING` transition. The later `QUEUED → FINALIZING` is then rejected as an invalid transition, so `tts_audio_end` is never emitted and the request queue stalls until timeout (log: `Invalid state transition … queued -> finalizing`). With a one-line fix — transition to `PROCESSING` when the dequeued item is still `QUEUED` — the test passes in ~30s. This affects every HTTP-TTS extension, so I've kept it out of this PR. Happy to open a separate `ten_ai_base` issue/PR.

Three issues the guarder surfaced **in this extension** are fixed in this branch:
- **`httpx[http2]` dependency** — the client uses `http2=True`; without the `h2` package it failed to initialize on every request.
- **`dump_path`** added to the `basic_audio_setting{1,2}` test configs (the guarder reads `config["dump_path"]`).
- **`mistral_tts_python` added to the fixed-sample-rate allowlist** in `tests/bin/start` (Voxtral emits a fixed 24 kHz, like `openai_tts2` / `humeai`).

wangyoucao577 (approved):

PR #2196, opened 2026-06-29
feat: enhance Soniox ASR with sentence termination handling

diyuyi-agora:


ai_agents/agents/ten_packages/extension/soniox_asr_python/text_utils.py
```diff
@@ -0,0 +1,176 @@
...
+            following_char
+        ):
+            return False
+
+        if tail == ".":
+            # Numeric periods need following-token context in the streaming
+            # path: "12.Hello" is a sentence boundary after "12.", while "12.5"
+            # is a decimal and must wait. If the next char is unknown ("12."),
+            # stay conservative and defer.
+            digit_period_decision = self._digit_period_boundary_decision(
```
wangyoucao577, line 160: pysbd can't handle this?

ai_agents/agents/ten_packages/extension/soniox_asr_python/extension.py
```diff
@@ -673,44 +744,112 @@ async def _process_transcript_and_translation(
...
+
+        await self._publish_asr_results(asr_results, translation_tokens)
 
-        for result in asr_results:
+    async def _publish_asr_results(
+        self,
+        asr_results: List[ASRResult],
+        translation_tokens: List[SonioxTranslationToken],
+    ) -> None:
+        """Publish one asr_result per segment, then asr_results batch."""
```
wangyoucao577: add some comments here since it's for backward compatibility.

ai_agents/agents/ten_packages/extension/soniox_asr_python/text_utils.py
diyuyi-agora: Because soniox is outputted in a streaming manner through tokens.
For both "12.Hello" and "12.5", it is possible to first receive the token "12.", and then receive the next token. When receiving "12.", only the currently existing token and the next token can be used for judgment.

Currently, pysbd only focuses on making judgments based on Mr., which is difficult to cover with rules

ai_agents/agents/ten_packages/extension/soniox_asr_python/extension.py
diyuyi-agora: DONE

zhangjie02agoraio (approved):

PR #2197, opened 2026-06-30
chore: add pyproject.toml files for python extensions

anexpn:
Add project metadata and dependency declarations for the Python extension packages under ai_agents/agents/ten_packages/extension.

Bump each extension manifest version to publish the packaging metadata updates.


anexpn:
Across the 94 added pyproject.toml files:

  - 64 are tighter than requirements.txt
  - 23 are identical
  - 1 is equivalent via tool.uv.sources (ten_vad_python)
  - 6 have no requirements.txt, so there’s nothing to compare
  - 0 are looser
  - 0 are mismatched in a way I’d call “different”

  What “tighter” means here:

  - Bare requirements like typing_extensions became lower-bounded entries like typing-extensions>=4.15.0
  - Bare pydantic became pydantic>=2.13.4
  - Some already-bounded lines stayed exactly the same, for example azure-cognitiveservices-speech==1.45.0
  - uv add resolved current minimum versions for unpinned packages, so the generated pyproject.toml is usually stricter than the original requirements.txt

  Example:

  - ai_agents/agents/ten_packages/extension/openai_asr_python/requirements.txt:1
  - ai_agents/agents/ten_packages/extension/openai_asr_python/pyproject.toml:1

  There:

  - typing_extensions became typing-extensions>=4.15.0
  - pytest became pytest>=9.1.1
  - pydantic became pydantic>=2.13.4
  - websockets>=14.0, openai>=1.99.3, and samplerate==0.1.0 stayed effectively the same

  Identical examples include:

  - ai_agents/agents/ten_packages/extension/websocket_server/requirements.txt:1
  - ai_agents/agents/ten_packages/extension/websocket_server/pyproject.toml:1

zhangjie02agoraio (approved):

PR #2203, opened 2026-07-06
feat: add smallest ai asr and tts extensions

harshitajain165:
## What

Adds two new vendor extensions for [Smallest AI](https://smallest.ai):

- **`smallest_asr_python`** — real-time speech-to-text using the
  [Pulse](https://docs.smallest.ai/waves/documentation/speech-to-text-pulse/overview)
  live WebSocket API (`wss://api.smallest.ai/waves/v1/stt/live`). 38 languages,
  64 ms time-to-first-transcript.
- **`smallest_tts_python`** — text-to-speech using the
  [Lightning](https://docs.smallest.ai/waves/documentation/text-to-speech-lightning/overview)
  SSE streaming endpoint (`/waves/v1/tts/live`). ~100 ms to first audio chunk,
  12 languages, voice cloning; `lightning_v3.1` and `lightning_v3.1_pro` models.

## Implementation notes

**ASR** (extends `AsyncASRBaseExtension`):
- Streams raw PCM16 binary frames; interim results with `final=false`, finals on
  Pulse's `is_final`.
- `asr_finalize` maps to Pulse's `{"type": "finalize"}` control message (session
  stays open).
- `word_timestamps=true` by default so final results carry accurate
  `start_ms`/`duration_ms`.
- Reconnect with exponential backoff (5 attempts) before reporting a fatal error.

**TTS** (extends `AsyncTTS2HttpExtension`):
- SSE frames (`data: {"audio": "<base64>"}` / `{"done": true}`) decoded
  incrementally; a partial line is never split across chunk boundaries.
- `output_format` pinned to `pcm`: Lightning's raw PCM is already signed 16-bit
  LE mono, matching the `pcm_frame` contract with no conversion, and skipping the
  container header keeps time-to-first-audio low.
- 401/403 and `invalid_api_key` classified as `INVALID_KEY_ERROR` (fatal);
  everything else non-fatal.

Both authenticate via `params.api_key` / `SMALLEST_API_KEY` and send
`X-Source: ten-framework` for API-side attribution. All other `params` keys pass
through verbatim (query string for ASR, request body for TTS).

## Testing

- Extension-local mocked test suites included for both (no API key needed):
  ASR — result shape, finalize, dump, metrics, reconnect, vendor error, invalid
  params; TTS — basic/dump/flush, error classification, metrics, params/URL
  resolution, robustness, state machine.
- `python -m black --check --line-length 80` clean; syntax verified.
- [x] `task asr-guarder-test EXTENSION=smallest_asr_python` — 10/10 passed against the live API
- [x] `task tts-guarder-test EXTENSION=smallest_tts_python` — 15/15 passed (+1 optional skip) against the live API
- [x] End-to-end voice test — live conversation via the playground on the `voice_assistant_smallest` graph (Pulse STT + GPT-4o + Lightning TTS over Agora RTC), including barge-in/interruption

_Supersedes #2200 (auto-closed by a force-push during a history cleanup)._


harshitajain165:
@halajohn @plutoless 
Requesting review whenever you find a chance, thanks in advance!

harshitajain165:
Hi @plutoless 👋

The earlier CI failure was in smallest_tts_python's test_headers_configuration (tests/test_params.py) — the client now sets a default X-Source: ten-framework header, so the header-count assertions were off by one. I've fixed the assertions and pushed the update.

The workflow runs are currently sitting in "action_required" (the fork-PR approval gate), so they haven't started. Could you approve/re-run the CI workflows when you get a chance? And a review whenever you have time would be much appreciated. Thanks a lot!

ai_agents/agents/ten_packages/extension/smallest_asr_python/extension.py
```diff
@@ -0,0 +1,561 @@
...
+            )
+            self.connected = False
+            await self.send_asr_error(
+                ModuleError(
+                    module=MODULE_NAME_ASR,
+                    code=ModuleErrorCode.NON_FATAL_ERROR.value,
+                    message=str(e),
+                ),
+            )
+            asyncio.create_task(self._handle_reconnect())
```
diyuyi-agora: When the vendor WebSocket closes, `_process_messages()` awaits `_handle_reconnect()` from inside the current `_message_task`.
  The reconnect path calls `start_connection()`, which first calls `stop_connection()`. `stop_connection()` then cancels the
  same `_message_task` that is currently executing the reconnect path, so the reconnect flow is interrupted before a new
  message-processing task is created.

plutoless: @harshitajain165 pls help fix

harshitajain165: @plutoless @diyuyi-agora 
Thanks for the detailed diagnosis, you were right about the self-cancellation.
I have added a fix. Every reconnect trigger now goes through a `_schedule_reconnect()` helper that runs the attempt on an independent, tracked task, so the message loop can exit cleanly while the reconnect proceeds. In-flight reconnect tasks are retained against GC and cancelled on `on_deinit`.
Please review

diyuyi-agora (dismissed):

diyuyi-agora:
format check failed， please fix

harshitajain165:
> format check failed， please fix

Could you re-run the CI workflows now?

plutoless (approved):

ranchangxin-agora:
Two issues need to be addressed:
1. Security: `SmallestTTSConfig.to_str(sensitive_handling=True)` masks only `params.api_key`, but `headers` are not redacted. `headers.Authorization` is a supported authentication path and is merged into the actual HTTP request headers. Since the full config is logged during initialization, an Authorization token can be written to the key-point log in plaintext. Redact `headers` with the shared `redact_headers()`/`redact_json()` helper before stringifying the config, and add coverage for Authorization and API-key header variants.
2. Connection status: `smallest_asr_python` never calls the base `on_connected()` after a successful WebSocket handshake, and does not call `on_disconnected()` on connection failure or socket close. The base class emits `CONNECTING` before `start_connection()`, so the reported `connection_status_changed` state remains `connecting` even when the socket is usable or has already failed. Call `on_connected()` immediately after `ws_connect()` succeeds, and call `on_disconnected()` with the available error/close details before reconnecting or stopping.

harshitajain165:
Thanks for the comment @ranchangxin-agora 
Have addressed both the issues in a separate PR: [link](https://github.com/TEN-framework/ten-framework/pull/2293/)

Please do review whenever you find a chance. Thanks in advance.

PR #2246, opened 2026-07-27
ci(docker): bake toolchains into build image, drop redundant CI installs

### LINKED ISSUE #917: Optimize the Docker image file used in the Linux environment of GitHub CI
Differentiate between the build-time and runtime Docker images, and update the toolchain inside the build-time Docker image so that there's no need to install toolchains separately in the GitHub CI steps.

noron12234:
Closes #917.

## Problem

Issue #917 asks to (1) differentiate build-time vs runtime Docker images and (2) bake toolchains into the build-time image so CI doesn't have to install them on every job.

Today, jobs that run inside `ghcr.io/ten-framework/ten_building_ubuntu2204` still spend 1-3 minutes per matrix cell re-installing tools that "should" already be in the image:

- `actions/setup-node@v4` runs 6× in `linux_ubuntu2204.yml` and 1× in `codeql.yml` — Node was never actually pre-installed in the image.
- `go install golang.org/dl/go1.24.3@latest` runs on every build.
- `rustup update nightly` runs on every debug build; `rustup default stable` runs on every release build. The image only ships the pinned `nightly-2025-09-18`.

On top of that, `ai_agents/Dockerfile` is already multi-stage but the build/runtime distinction is not obvious to a first-time reader.

## Fix

**`tools/docker_for_building/ubuntu/22.04/Dockerfile`**
- Add Node.js 20 (LTS) via NodeSource.
- Pre-install Go 1.24.3 alongside the existing 1.20.12 and 1.22.3.
- Install `stable` and latest `nightly` Rust toolchains alongside the pinned `nightly-2025-09-18` default.

**`.github/workflows/linux_ubuntu2204.yml`**
- Drop 6 redundant `actions/setup-node@v4` steps in container-based jobs. Keep the one in `test-integration-python` (runs on a bare `ubuntu-22.04` runner, not the container).
- Drop redundant `go install golang.org/dl/go1.24.3@latest` and `rustup update nightly` in the `build` job.

**`.github/workflows/codeql.yml`**
- Drop redundant `actions/setup-node@v4`.

**`ai_agents/Dockerfile`**
- Header comments only, making the build-time vs runtime stages explicit. No functional change; runtime image bytes are identical.

## Verification

- YAML syntax on both edited workflows: OK (`yaml.safe_load`).
- Hadolint on both Dockerfiles: no new warnings.
- NodeSource install on \`ubuntu:22.04\`: verified in isolated container (Node 20.20.2, npm 10.8.2 installed).
- Rustup multi-toolchain install (`--default-toolchain nightly-XXXX` plus `toolchain install stable` plus `toolchain install nightly`): verified in isolated container, `rustup default` switches cleanly.
- Full local `docker build` of the build-time image was attempted but is a 30+ min job; the definitive rebuild+publish happens in CI via `.github/workflows/build_docker.yml` on merge.

## Why this way (vs alternatives)

- **Bake into the image, not into a composite action**: composite actions still add per-job wall time and network I/O. The image is cached on the runner, so per-job cost approaches zero.
- **Kept `test-integration-python`'s `setup-node`**: that job runs on a bare runner (`ubuntu-22.04`) for disk-space reasons. Removing it there would break Node.
- **Added `stable` + latest `nightly` alongside pinned nightly**: keeps the pinned nightly as default (unchanged behavior), and lets release/debug matrix cells switch via `rustup default` alone with zero network I/O.
- **Left `ai_agents/Dockerfile` runtime `apt-get` list untouched**: `python3-pip` / `python3-dev` / `python3-venv` are technically build tools, but downstream Python extensions may install packages at container run time. Removing them is a separate risk I didn't want to bundle with this PR.

## Trade-offs

Image size will grow from ~3.9 GB to ~5.5 GB (Rust extras + Node). Net CI cost is a win: every job saves 1-3 min setup time × N matrix cells; concrete savings will show in the CI on this PR once `build_docker.yml` rebuilds and republishes the image to ghcr.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -114,10 +114,7 @@ jobs:
             git submodule update --init --recursive --depth 1 "$sub"
           done
 
-      - name: Setup Node.js
-        uses: actions/setup-node@v4
-        with:
-          node-version: 20
+      # Node.js is pre-installed in ten_building_ubuntu2204 (see issue #917).
 
       - name: Initialize CodeQL
         uses: github/codeql-action/init@v3
--- a/.github/workflows/linux_ubuntu2204.yml
+++ b/.github/workflows/linux_ubuntu2204.yml
@@ -126,9 +126,7 @@ jobs:
             git submodule update --init --recursive --depth 1 "$submodule"
           done
 
-      - uses: actions/setup-node@v4
-        with:
-          node-version: 20
+      # Node.js is pre-installed in ten_building_ubuntu2204 (see issue #917).
 
       - name: Update version
         run: |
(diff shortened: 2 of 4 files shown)


noron12234:

PR #2263, opened 2026-07-29
fix: apply pre-flatten passes to imported subgraphs

noron12234:
## Problem

`Graph::flatten_graph` runs four passes over a graph:

1. `expand_names_to_individual_items` — `names: [...]` into individual `name` items
2. `flatten_selectors` — selector references into the nodes they match
3. `convert_reversed_connections_to_forward_connections` — `source` into `dest`
4. `flatten_subgraphs` — inline imported subgraphs, prefixing names with the subgraph name

Passes 1-3 were only ever applied to the top-level graph. `process_subgraph_node` loads an imported subgraph with `load_graph_from_uri`, which only deserializes the file, and hands it straight to pass 4. A subgraph that uses `names`, a `selector` node, or a reversed `source` connection therefore reaches the flattening logic unlowered, even though that logic assumes all three passes have already run.

## Reproduction

`sub.json` — inside the subgraph, `ext_a` sends `hello` to `ext_b`, written in reverse form:

```json
{
  "nodes": [
    {"type": "extension", "name": "ext_a", "addon": "addon_a"},
    {"type": "extension", "name": "ext_b", "addon": "addon_b"}
  ],
  "connections": [
    {"extension": "ext_b", "cmd": [{"name": "hello", "source": [{"extension": "ext_a"}]}]}
  ]
}
```

Main graph:

```json
{"nodes": [{"type": "subgraph", "name": "sg", "graph": {"import_uri": "file://.../sub.json"}}]}
```

Expected after flattening: `sg_ext_a --hello--> sg_ext_b`.

Actual on `main`:

```json
{
  "nodes": [
    {"type": "extension", "name": "sg_ext_a", "addon": "addon_a"},
    {"type": "extension", "name": "sg_ext_b", "addon": "addon_b"}
  ],
  "connections": [
    {"extension": "sg_ext_b", "cmd": [{"name": "hello", "source": [{"extension": "ext_a"}]}]}
  ]
}
```

The connection is still in reverse form, and its `source` still names `ext_a` while the node has been renamed to `sg_ext_a`. It now points at an extension that does not exist, and nothing reports an error.

## Impact per feature

| Written inside an imported subgraph | Result on `main` |
| --- | --- |
| `source` (reversed connection) | Silently produces a connection referencing a non-existent extension. No error, message never delivered. |
| `names: [...]` | Reaches flattening unexpanded, where `name` is expected to be `Some` and `names` to be `None`. |
| `selector` node | Panics: `Unexpected non-extension node in flattened subgraph` (`subgraph/flatten.rs:493`). |

The `selector` case is the most visible one: this code path is reachable from the tman designer HTTP backend and from the C runtime through `ten_rust_graph_validate_complete_flatten`, so a malformed-but-plausible subgraph file takes down the process instead of returning an error.

## Fix

Extract passes 1-3 into `Graph::apply_pre_flatten_passes` and run them on a subgraph immediately after it is loaded, so an imported subgraph goes through the same lowering as the top-level graph. `flatten_graph` now calls the same helper, so there is a single definition of what has to happen before flattening.

The other `load_graph_from_uri` caller (`GraphContent::validate_and_complete_and_flatten`) already reaches these passes through `flatten_graph` and needed no change.

## Tests

Three tests added to `tests/test_case/graph/subgraph.rs`, one per feature:

- `test_flatten_subgraph_with_reversed_connection`
- `test_flatten_subgraph_with_names_array`
- `test_flatten_subgraph_with_selector`

Verified they fail on `main` and pass with this change:

```
# without the fix
test test_flatten_subgraph_with_names_array ... FAILED
test test_flatten_subgraph_with_reversed_connection ... FAILED
test test_flatten_subgraph_with_selector ... FAILED     # panics at flatten.rs:493
test result: FAILED. 9 passed; 3 failed

# with the fix
test result: ok. 12 passed; 0 failed
```

The rest of the graph suite is unaffected — `graph::reverse` (14), `graph::flatten_integration` (4), `graph::names_expansion` (6), `graph::selector` (2), `graph::exposed_message` (10), `graph::graph_info` (5), `graph::import_uri` (6) all still pass.

## Possible follow-up

`validate_and_complete` is also never run on an imported subgraph, so the app-URI rules are not enforced there. I left that out of this PR because it changes which graphs are accepted rather than fixing a mis-compilation. Happy to send a separate PR if you would like that tightened too.

### DIFF
--- a/core/src/ten_rust/src/graph/mod.rs
+++ b/core/src/ten_rust/src/graph/mod.rs
@@ -526,6 +526,39 @@ impl Graph {
         Ok(Some(new_graph))
     }
 
+    /// Applies the passes that must run before subgraph flattening:
+    ///
+    /// 1. Expand `names` arrays into individual `name` items.
+    /// 2. Replace selector references with the nodes they match.
+    /// 3. Convert reversed connections (`source`) to forward connections
+    ///    (`dest`).
+    ///
+    /// `flatten_subgraphs` relies on all three having already run - it expects
+    /// `name` to be `Some`, `names` to be `None`, and every node to be an
+    /// extension node. Any graph handed to it, including one imported through
+    /// `import_uri`, has to go through here first.
+    ///
+    /// Returns `Ok(None)` if none of the passes changed anything.
+    pub fn apply_pre_flatten_passes(&self) -> Result<Option<Graph>> {
+        let mut processing_graph = self;
+
+        let expanded_names_graph = processing_graph.expand_names_to_individual_items()?;
+        processing_graph = expanded_names_graph.as_ref().unwrap_or(processing_graph);
+
+        let flattened_selector_graph = processing_graph.flatten_selectors()?;
+        processing_graph = flattened_selector_graph.as_ref().unwrap_or(processing_graph);
+
+        let reversed_graph =
+            processing_graph.convert_reversed_connections_to_forward_connections()?;
+        processing_graph = reversed_graph.as_ref().unwrap_or(processing_graph);
+
+        if std::ptr::eq(processing_graph, self) {
+            return Ok(None);
+        }
+
+        Ok(Some(processing_graph.clone()))
+    }
+
     /// Convenience method for flattening a graph instance without preserving
     /// exposed info. This is the main public API for flattening graphs.
     ///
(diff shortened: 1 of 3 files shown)


noron12234:
Closing for now — I want to re-check this myself before asking for review. Will reopen once I'm confident. Sorry for the noise.

noron12234:
Reopened and rebased onto current `main`.

Verified the bug is still present. Applying only the new tests to `main`, without the fix:

- `test_flatten_subgraph_with_names_array` — fails, `names` reaches flattening unexpanded
- `test_flatten_subgraph_with_reversed_connection` — fails, `source` still references `ext_a` after the node was renamed to `sg_ext_b`
- `test_flatten_subgraph_with_selector` — panics at `subgraph/flatten.rs:493`

With the fix, all 87 tests under `graph::` pass.

On the earlier CI failure: it failed at `Setup tgn`, not in the Rust code. The same workflow also failed on the `cleanup` branch around the same time, and #2269 (rust toolchain replacement cleanup) has since landed. This branch now includes it.

PR #2264, opened 2026-07-30
fix(tts): correct cosy audio end interval

YiminW:


wangyoucao577 (dismissed):

YiminW:
1. on_error() will set done which aovids duplicate error msgs
2. not necessary
3. this pr only fix cosy tts issue

wangyimin-agora:
## CosyVoice hardening update

The branch now contains the complete repository-owned CosyVoice hardening change, split into focused commits so the functional and test diffs are easy to review.

### Runtime

- Keep the official DashScope SDK, with a preconnected `SpeechSynthesizerObjectPool` shared by extension instances in one worker.
- Return a synthesizer only after synchronous task completion; close and replace failed connections.
- Add bounded pool wait, first-audio, task, input-idle, and cancellation timeouts.
- Correct request lifecycle handling for completion, flush/cancel, stale callbacks, provider errors, and recovery on the next request.
- Preserve streaming multi-chunk text input. `enable_ssml=true` is rejected because the provider only accepts SSML as a single text chunk.
- Forward unknown `params` keys to DashScope so newly introduced provider parameters work without extension changes.
- Support workspace ID, custom headers, custom endpoint, and all supported PCM sample rates.
- Fix audio-end interval semantics, dump behavior, usage calculation, structured error reporting, connection status, connect-delay, and TTFB metrics.

### Diagnostics and tests

- Add matched SDK and direct-WebSocket TTFB benchmark scripts for isolating SDK overhead.
- Add deterministic coverage for parameter forwarding, object-pool setup, dump/flush, metrics, provider errors, recovery, and request state transitions.
- Validation: Cosy standalone suite `11 passed`; target Pylint `10.00/10`; Black and `git diff --check` passed.

### Current limits

- DashScope SDK remains the production transport; direct WebSocket is diagnostic only.
- Output remains mono 16-bit PCM.
- Word-level timestamps are not exposed yet.
- Streaming SSML is intentionally unsupported; normal text remains streaming.

PR #2277, opened 2026-08-04
fix: update inworld tts to current api contract

BenWeekes:
## What

The `inworld_tts_python` extension is currently silent for everyone: the Inworld TTS API has drifted since the extension landed (#2010) and every synthesis request now fails.

Changes to `inworld_tts.py`:
- Request fields renamed: `voiceId`/`modelId` (the previous `voice`/`model` are rejected with `400: voice_id can not be empty`), and output format moved to `audio_config` (`audio_encoding`, `sample_rate_hertz`, verified honored at 16k/24k).
- Streamed audio is now nested under `result.audioContent` in each NDJSON line; the old top-level paths are kept as fallbacks.
- Each streamed chunk is a self-contained WAV; the RIFF header is stripped per chunk so raw PCM reaches the pipeline (previously this would have produced audible clicks).
- Bad credentials now surface as HTTP 403 ("Invalid authorization credentials"); 401/403 are both treated as fatal invalid-key errors so the guarder `-1000` contract holds.

## Tests

- New standalone suite under `tests/` (dump, flush, params passthrough, error contracts incl. invalid key, TTFB metrics, robustness, state machine): **9 passed**.
- New `tests/configs` for the TTS guarder; full guarder run against the live API: **15 passed, 2 skipped** (subtitle alignment, disabled by default).

All verified end-to-end in a voice-assistant graph (deepgram asr → openai llm → inworld tts) with audible output.


BenWeekes:
Guarder test result (run against the live Inworld API):

```
$ task tts-guarder-test EXTENSION=inworld_tts_python
================== 15 passed, 2 skipped in 345.05s (0:05:45) ===================
```

The 2 skips are the subtitle-alignment tests, disabled by default for all TTS extensions.

Standalone extension suite:

```
$ task test-extension EXTENSION=agents/ten_packages/extension/inworld_tts_python
============================== 9 passed in 1.28s ===============================
```

Version bumped to 0.1.2 in both `manifest.json` and `pyproject.toml`.

BenWeekes:
Guarder test result (verbose run against the live Inworld API):

tts guarder result

PR #2278, opened 2026-08-04
fix: migrate openai mllm to ga realtime api and add per-join greeting

BenWeekes:
## What

Two changes to `openai_mllm_python`:

**1. Migrate to the GA Realtime API shape.** OpenAI retired the beta protocol (`beta_api_shape_disabled`), which currently breaks this extension for everyone — sessions cannot even be created. This drops the `OpenAI-Beta` header, maps the renamed GA server events onto the existing message classes (`response.output_audio.delta`, `response.output_audio_transcript.delta`, `conversation.item.added/done`, …), and sends the GA `session.update` shape (`type: realtime`, `output_modalities`, nested `audio.input/output` with transcription and `turn_detection`).

**2. Per-join greeting param.** `greeting` config field (default `""` = off), spoken once per join after the session config is applied, as a user message item plus a response request — the same semantics as `gemini_mllm_python`'s existing `greeting`. This makes the extension compatible with platforms that supply a greeting per join instead of hardcoding it in graph properties.

## Verification

- Verified against the live GA service: session created and updated, audio in/out, transcripts, and spoken greeting all working in a voice-assistant graph.
- The greeting delivery path (item.create + response.create) is the same code path main_control already exercises in the realtime example.

### DIFF
--- a/ai_agents/agents/ten_packages/extension/openai_mllm_python/extension.py
+++ b/ai_agents/agents/ten_packages/extension/openai_mllm_python/extension.py
@@ -57,14 +57,9 @@ from .realtime.struct import (
     InputAudioBufferSpeechStopped,
     ResponseFunctionCallArgumentsDone,
     ErrorMessage,
-    SessionUpdate,
-    SessionUpdateParams,
-    InputAudioTranscription,
     ContentType,
     FunctionCallOutputItemParam,
     ResponseCreate,
-    ServerVADUpdateParams,
-    SemanticVADUpdateParams,
 )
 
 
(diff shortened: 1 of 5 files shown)


BenWeekes:
Version bumped to 0.2.3 in both `manifest.json` and `pyproject.toml`.

No guarder suite exists for MLLM extensions (only `asr_guarder`/`tts_guarder`); verification was done against the live GA Realtime service in a voice-assistant graph: session create/update, bidirectional audio, transcripts, and the per-join greeting all confirmed working.

PR #2306, opened 2026-08-31
fix(asr): improve connection handling and add locking for audio sending

diyuyi-agora:
修复 update_configs 重连与音频发送并发时，连接被中途关闭并触发 ASR -1000 Not connected 的问题。通过串行化音频发送和连接切换，确保在途音频发送完成后再重连。

### DIFF
--- a/ai_agents/agents/ten_packages/extension/bytedance_llm_based_asr/extension.py
+++ b/ai_agents/agents/ten_packages/extension/bytedance_llm_based_asr/extension.py
@@ -175,8 +175,12 @@ class BytedanceASRLLMExtension(AsyncASRBaseExtension):
         self.enable_utterance_grouping: bool = True
 
         self._update_configs_lock: asyncio.Lock = asyncio.Lock()
+        # Serialises the whole stop+start swap (update_configs vs reconnect).
+        # Acquire before _send_lock when both are needed.
+        self._swap_lock: asyncio.Lock = asyncio.Lock()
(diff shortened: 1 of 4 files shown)


ai_agents/agents/ten_packages/extension/bytedance_llm_based_asr/extension.py
```diff
@@ -413,78 +450,92 @@ def is_connected(self) -> bool:
...
-            # Dump audio to log_id_dumper if enabled (manager handles rename if needed)
             if self.log_id_dumper_manager:
                 await self.log_id_dumper_manager.push_bytes(audio_data)
 
-            self.audio_timeline.add_user_audio(
-                int(len(buf) / (self.input_audio_sample_rate() / 1000 * 2))
-            )
+            async with self._send_lock:
+                client = self.client
+                if not self.is_connected() or client is None:
```
diyuyi-agora, line 490: `send_audio()` can return `False` after `_handle_audio_frame()` has already passed its initial connection check, for example when a connection swap starts while awaiting the audio dumper, log-id dumper, or `_send_lock`. `AsyncASRBaseExtension` only buffers before calling `send_audio()` and does not requeue a frame when the call returns `False`; it also ignores failures while flushing `buffered_frames`. This can silently drop the current frame, or multiple buffered frames. Please retain/requeue the frame when the send fails, or make the connection-check-to-send path atomic.

PR #2307, opened 2026-09-01
feat: speko.ai router extensions

idafoh:


ai_agents/agents/ten_packages/extension/speko_tts2_python/extension.py
```diff
@@ -0,0 +1,360 @@
...
+
+    def vendor(self) -> str:
+        return "speko"
+
+    def vendor_metadata(self) -> dict[str, Any]:
+        if self.config is None:
+            return {}
+        metadata: dict[str, Any] = {
+            "base_url": self.config.base_url,
+            "routing": self.config.routing,
```
wangyimin-agora, line 104: "key": 
            "base_url": 
            "language": 
            "api_key": 
            "routing":

key and api_key have same vaule. you can refer to TEN-Agent/ai_agents/agents/ten_packages/extension/rime_tts/extension.py

ai_agents/agents/ten_packages/extension/speko_tts2_python/extension.py
```diff
@@ -0,0 +1,360 @@
...
+                    "Ignoring text for a completed Speko TTS request"
+                )
+                return
+
+            text = text_input.text.strip()
+            if text:
+                await self._stream_text(text, text_input.request_id)
+
+            if text_input.text_input_end:
+                await self._finalize_request(TTSAudioEndReason.REQUEST_END)
```
wangyimin-agora: _finalize_request only be called when receiving all audio of the same reqeust id.
when text_input.text_input_end is true, we receivesall text of the same reqeust id from llm, and also need to wait for audio from tts server

ai_agents/agents/ten_packages/extension/speko_tts2_python/extension.py
```diff
@@ -0,0 +1,360 @@
...
+        except Exception as error:
+            self.ten_env.log_error(
+                f"Speko TTS request failed: {traceback.format_exc()}"
+            )
+            await self.on_disconnected(
+                code=ModuleErrorCode.NON_FATAL_ERROR.value,
+                message=str(error),
+                vendor_info=ModuleErrorVendorInfo(vendor=self.vendor()),
+            )
+            await self._finalize_request(
```
wangyimin-agora: excpet may happen during request id, so we need to add condition before call this funtion.
you can refer to TEN-Agent/ai_agents/agents/ten_packages/extension/rime_tts/extension.py

wangyimin-agora: This exception may occur while processing a non-final text chunk (text_input_end=False). Calling _finalize_request() here prematurely completes the whole request, so the remaining chunks with the same request ID can no longer be processed. Please only finalize when the request has received its final input; for an intermediate-chunk error, report the error and reset/reconnect the client without calling finish_request(), similar to the Rime TTS implementation.

ai_agents/agents/ten_packages/extension/speko_tts2_python/extension.py
```diff
@@ -0,0 +1,360 @@
...
+        self._finalized = True
+        await self._ensure_audio_start()
+        await self.send_tts_audio_end(
+            request_id=request_id,
+            request_event_interval_ms=self._request_interval_ms(),
+            request_total_audio_duration_ms=self._audio_duration_ms(),
+            reason=reason,
+            extra_metadata={"router_usage": self._router_usage},
+        )
+        await self._close_client()
```
wangyimin-agora: client.close() can raise an error while draining the final session.close response, as demonstrated by test_close_surfaces_terminal_router_error. If that happens here, finish_request() below is skipped. Since _finalized has already been set to True, the outer error handler only emits an error and does not complete the request state, leaving _processing_request_id occupied and potentially blocking all subsequent requests. Please ensure finish_request() and recorder cleanup always run exactly once, even when closing the vendor session fails. Also avoid emitting a successful tts_audio_end before handling a close-time error.

ai_agents/agents/ten_packages/extension/speko_tts2_python/extension.py
```diff
@@ -0,0 +1,360 @@
...
+        )
+        self.ten_env.log_info(
+            f"vendor_status_changed: Speko TTS ready; route={self._route}",
+            category=LOG_CATEGORY_VENDOR,
+        )
+        return client
+
+    async def _stream_text(self, text: str, request_id: str) -> None:
+        client = await self._ensure_client()
+        self.metrics_add_input_characters(len(text))
```
wangyimin-agora: Please use metrics_add_output_characters(len(text)) here. TTS extensions, including Rime TTS and the HTTP TTS base implementation, count text submitted for synthesis as output characters. The current implementation records it as input characters, so output_characters remains zero for every Speko request and the reported usage metrics are inconsistent with other TTS extensions.

wangyimin-agora (changes requested):
please change code as comment and run tts guarder test for this new extension

ai_agents/agents/ten_packages/extension/speko_asr_python/client.py
```diff
@@ -0,0 +1,256 @@
...
+            return
+
+        self._finalize_waiter = asyncio.get_running_loop().create_future()
+        await self._ws.send('{"type":"input.commit"}')
+        try:
+            await asyncio.wait_for(
+                asyncio.shield(self._finalize_waiter),
+                timeout=self.finalize_timeout_sec,
+            )
+        except asyncio.TimeoutError as error:
```
diyuyi-agora, line 177: **Blocking:** On `asyncio.TimeoutError`, `_finalize_waiter` remains pending. The next `commit()` hits:

```python
if self._finalize_waiter and not self._finalize_waiter.done():
    await self._finalize_waiter
    return
```

and can block forever. Please fail/cancel/reset the waiter on timeout (and consider the auto-final-before-`asr_finalize` case from the guarder tests).

ai_agents/agents/ten_packages/extension/speko_asr_python/extension.py
```diff
@@ -0,0 +1,359 @@
...
+
+    @override
+    def vendor(self) -> str:
+        return "speko"
+
+    @override
+    def vendor_metadata(self) -> dict[str, Any]:
+        if self.config is None:
+            return {}
+        metadata: dict[str, Any] = {
```
diyuyi-agora, line 57: **Blocking:** `vendor_metadata()` should expose `key` (from `api_key`), similar to `azure_asr_python` / `tencent_asr_python` and the `rime_tts` pattern referenced in the TTS review. Consider also including `language` and omitting empty values.

diyuyi-agora (commented):
## Review: `speko_asr_python`

Thanks for adding the Speko Router ASR extension — the overall structure looks good and follows existing TEN ASR patterns (`AsyncASRBaseExtension`, separated client/config, guarder-ready property configs, client unit tests).

Below are findings focused on the ASR extension. (TTS feedback is tracked separately.)

### Blocking

1. **`commit()` timeout leaves a pending `_finalize_waiter`** — after a timeout, the future is never failed/cleared, so a subsequent `commit()` can block forever on `await self._finalize_waiter`. This can also happen when the vendor auto-finalizes before `asr_finalize` (covered by `test_same_session_finalize_reconnect`).

2. **`vendor_metadata()` is missing `key`** — please align with other ASR extensions (`azure_asr_python`, `tencent_asr_python`) and the `rime_tts` convention already referenced in this PR.

### Should fix

3. **No `on_deinit()` cleanup** for `audio_dumper` (only stopped in `stop_connection()`).
4. **`session.closed` usage is stored but never forwarded** to the extension/metrics path.
5. **`send_audio()` errors only return `False`** — consider disconnect/reconnect instead of leaving a half-dead session.
6. **Per-turn timestamp state is not reset** — `_total_audio_bytes` / `_final_cursor_ms` accumulate across multiple finalize cycles in the same session, which can break timestamp accuracy when `segments` is absent.
7. **Auto-final before `asr_finalize` may emit a spurious timeout error** even though `asr_finalize_end` is still sent.
8. **Missing extension-level / guarder coverage** — only client tests exist today; please add README instructions and run:

```bash
cd ai_agents/agents/ten_packages/extension/speko_asr_python/tests && ./bin/start
cd ai_agents && task asr-guarder-test EXTENSION=speko_asr_python
```

Pay special attention to `test_asr_finalize`, `test_same_session_finalize_reconnect`, `test_metrics`, and `test_audio_timestamp`.

Happy to re-review once these are addressed.

ai_agents/agents/ten_packages/extension/speko_asr_python/client.py
```diff
@@ -0,0 +1,256 @@
...
+                    )
+                event = self._decode_text_event(raw)
+                event_type = event.get("type")
+                if event_type == "error":
+                    error = SpekoRouterError.from_event(event)
+                    break
+                if event_type == "usage.updated":
+                    self.usage = dict(event.get("usage", {}))
+                elif event_type == "session.closed":
+                    self.usage = dict(event.get("usage", {}))
```
diyuyi-agora, line 229: On `session.closed`, usage is saved to `self.usage` but `on_event()` is not called before `break`, so the extension never receives the final usage metrics. Please forward this (either emit a synthetic `usage.updated` or report usage during disconnect).

ai_agents/agents/ten_packages/extension/speko_asr_python/extension.py
```diff
@@ -0,0 +1,359 @@
...
+            await self.client.send_audio(audio)
+            self._total_audio_bytes += len(audio)
+            self.audio_timeline.add_user_audio(
+                self._audio_duration_ms(len(audio))
+            )
+            if self.audio_dumper is not None:
+                await self.audio_dumper.push_bytes(audio)
+            return True
+        except SpekoRouterError as error:
+            await self._report_router_error(error)
```
diyuyi-agora: On `SpekoRouterError` during `send_audio`, the method only reports the error and returns `False`. Other ASR extensions typically disconnect/reconnect here to avoid silently dropping audio on a broken session.

ai_agents/agents/ten_packages/extension/speko_asr_python/extension.py
```diff
@@ -0,0 +1,359 @@
...
+                final=final,
+                start_ms=start_ms,
+                duration_ms=max(0, end_ms - start_ms),
+                language=self.config.language if self.config else "en-US",
+                words=words,
+                metadata=metadata,
+            )
+        )
+        if final:
+            self._final_cursor_ms = max(self._final_cursor_ms, end_ms)
```
diyuyi-agora, line 399: `_final_cursor_ms` / `_total_audio_bytes` are never reset between turns in the same WebSocket session. If a transcript arrives without `segments`, later turns can get incorrect `start_ms` / `duration_ms` (relevant for `test_audio_timestamp` and multi-turn guarder tests).

ai_agents/agents/ten_packages/extension/speko_asr_python/README.md
```diff
@@ -0,0 +1,43 @@
...
+    "params": {
+      "api_key": "${env:SPEKO_API_KEY}",
+      "routing": {"mode": "auto", "objective": "latency"},
+      "language": "en-US"
+    }
+  }
+}
+```
+
+For an explicit route, set `routing` to
```
diyuyi-agora, line 42: Please add a Testing section with the guarder command, similar to `smallest_asr_python`:

```bash
task asr-guarder-test EXTENSION=speko_asr_python
```

Also consider adding `test_package_contract.py` and extension-level finalize/metrics tests.

idafoh:
@wangyimin-agora @diyuyi-agora thanks for the reviews! I've pushed updates addressing the ASR/TTS feedback, added regression tests, and synced with the latest main. Could you please take another look when you have a chance?

ai_agents/agents/ten_packages/extension/speko_tts2_python/extension.py
```diff
@@ -0,0 +1,448 @@
...
+                    )
+
+    def _make_module_error(self, error: SpekoRouterError) -> ModuleError:
+        if not error.retryable and error.code in {
+            "authentication_failed",
+            "insufficient_credit",
+            "route_not_found",
+            "capability_unsupported",
+        }:
+            self._permanent_error = error
```
wangyimin-agora, line 340: I think  _permanent_error means fatal error, which cannot recover, but _module_error_code is reported as non fatal error

wangyimin-agora (changes requested):
1. please provide Screenshot of tts guarder test result
2. fix error type of  _module_error_code()
3. fix failed CI jobs. For format/lint issue, you can run task lint in ten framewwork image container

diyuyi-agora:
Rebase the main branch and push it again. Have AI review it again.

ai_agents/agents/ten_packages/extension/speko_tts2_python/extension.py
idafoh: You’re right, that was inconsistent. Fixed in both ASR and TTS: permanent errors now report fatal severity and prevent further connection attempts.

idafoh:
> please provide Screenshot of tts guarder test result

Screenshot 2026-09-18 at 2 49 33 PM
Screenshot 2026-09-18 at 3 00 29 PM

Here are the test results

YiminW:
great， no more comments for tts

PR #2311, opened 2026-09-07
fix: fix fish audio tts flush issue

YiminW:


wangyimin-agora:
1. non-final chunk shoud not send audio end
2. base tts ensure these behaviour
3. align design
4. align design
5. no issue
6. base class function is verified by ut
7. not this PR scope

wangyoucao577 (approved):

PR #2320, opened 2026-09-14
feat(asr): add rtzr streaming extension

kimdwkimdw:
## Summary

Add `rtzr_asr_python` to make RTZR streaming speech recognition available through TEN's standard ASR interface. The extension follows `AsyncASRBaseExtension` and uses `aiohttp` directly, without an additional vendor SDK dependency.

- Support `sommers_ko`, `sommers_ja`, and `sommers_en`, mapping interim/final transcripts, timestamps, and word timing to TEN results.
- Implement authentication and token reuse, same-connection Finalize, EOS shutdown, bounded audio buffering and retries, connection status, metrics, and optional audio dumping.
- Configure credentials and endpoints through `RTZR_CLIENT_ID`, `RTZR_CLIENT_SECRET`, `RTZR_API_BASE`, and `RTZR_WEBSOCKET_URL`. Default endpoints are `https://openapi.vito.ai` and `wss://openapi.vito.ai`.
- Add standalone tests, opt-in live tests, test configurations, README setup/graph examples, and registration for the common ASR guarder's connection-status tests.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

## Testing

- [x] Tests added/updated
- [ ] All repository tests pass — the full repository suite was not run
- [x] Manual testing completed against an existing RTZR development service

| Check | Result |
| --- | --- |
| Final standalone offline suite, Python 3.10 / pytest 8 | 39 passed |
| Final common ASR guarder | 11 passed, 2 deselected |
| Final Japanese live recognition with repeated Finalize | Passed with 3 CSJ utterances |
| Three-model continuous streaming | 300 seconds per model, no errors; run before the final malformed-response validation change |
| Direct API vs. TEN transcripts | Normalized transcripts matched for the Korean, Japanese, and English samples, including the CSJ follow-up |
| Ruff / Black / Pylint | Passed; Pylint 10.00/10 |
| TEN manifest/property schemas and package creation | Passed |

The common guarder's Chinese-specific multilingual case was replaced by model-specific Korean, Japanese, and English live tests. Its long-duration case was replaced by the 300-second-per-model tests. These live checks use small samples and do not establish corpus-wide recognition quality.

Live tests are opt-in and require credentials and an external audio manifest. Credentials, evaluation audio, reference transcripts, and local experiment artifacts are not included in this change. No upstream CI result is claimed.

## Documentation

- [x] Documentation updated
- [x] Examples provided if needed

The extension README covers dependencies, environment/property configuration, graph wiring, lifecycle behavior, and test commands. The shared `.env.example` includes the RTZR variables.

## Breaking Changes

None. This is an opt-in extension and does not change existing ASR implementations or application graphs.

### Protocol limitations

- RTZR does not send a separate Finalize acknowledgement for silence following an automatic final. When no hypothesis remains pending, the extension reports local completion after a bounded wait with a `finalize_without_result` metric; it does not fabricate a vendor final result.
- Audio already sent before a disconnect is not replayed because the streaming protocol provides no per-frame acknowledgement/resume mechanism.


ai_agents/agents/ten_packages/extension/rtzr_asr_python/extension.py
```diff
@@ -0,0 +1,455 @@
...
+                    text=text,
+                    final=payload["final"],
+                    start_ms=offset,
+                    duration_ms=duration,
+                    language=self.config.language,
+                    words=words,
+                )
+            )
+        if payload["final"]:
+            self._final_bytes = self._sent_bytes
```
diyuyi-agora: [P1] Correlate completion with the finalized audio boundary

The vendor result's `start_at`/`duration` describes the audio it finalized, but this assigns every byte sent by callback time. A delayed automatic-final can arrive after frames for the next utterance have already been sent; `_final_bytes` then jumps to `_sent_bytes`, and `finalize()` skips waiting at line 366 and emits `asr_finalize_end` while those later frames are still unfinalized. I reproduced this with 1 second sent and a delayed final covering only the first 500 ms: the subsequent finalize immediately emitted only `asr_finalize_end`. Please capture the target byte/time boundary when `Finalize` is sent and only satisfy it when a final result reaches that boundary (or correlate by sequence/turn), with a regression test for delayed auto-final plus trailing audio.

ai_agents/agents/ten_packages/extension/rtzr_asr_python/extension.py
```diff
@@ -0,0 +1,455 @@
...
+                self._received_result = False
+                self._receiver = asyncio.create_task(self._receive(self.ws))
+                await self.on_connected()
+                await self.send_connect_delay_metrics(
+                    max(1, int((time.monotonic() - start) * 1000))
+                )
+                await self._flush_frames()
+            except Exception as exc:
+                status = getattr(exc, "status", None)
+                fatal = status in (400, 401, 403, 404) or isinstance(
```
diyuyi-agora, line 138: [P2] Retry a WebSocket 401 after invalidating the cached token

`RTZRClient.connect()` clears its cached token when the WebSocket handshake returns 401, but this classifies every 401 as fatal, so `_schedule_retry()` becomes a no-op and the cleared token is never refreshed. If a cached token is revoked before `expire_at`, otherwise-valid client credentials cannot recover. Please distinguish an authentication-endpoint 401 from a WebSocket cached-token 401, or retry the handshake once after forcing token refresh; only invalid client credentials should be terminal.

ai_agents/agents/ten_packages/extension/rtzr_asr_python/config.py
```diff
@@ -0,0 +1,98 @@
+import json
+from typing import Any
+from urllib.parse import urlsplit, urlunsplit
+
+from pydantic import BaseModel, Field, model_validator
+
+MODEL_LANGUAGES = {
+    "sommers_ko": "ko-KR",
+    "sommers_ja": "ja-JP",
+    "sommers_en": "en-US",
```
diyuyi-agora: [P2] Align the English model with the configured public endpoint

The current public RTZR streaming documentation lists `sommers_ko`, `sommers_ja`, and `whisper`, but not `sommers_en`: https://developers.rtzr.ai/docs/en/stt-streaming/. This extension defaults to the public `openapi.vito.ai` endpoint, while `property_en.json`, dump/invalid configs, and the live matrix use `sommers_en`; the PR description also says live validation used a development service. As submitted, a normal public-endpoint English guarder run is likely to fail the handshake with invalid parameters. Please either use the public `whisper` model with an English language parameter, publish/support `sommers_en` on the public endpoint before merge, or make the non-public endpoint requirement explicit and avoid using this model in the default public configs.

diyuyi-agora (commented):
Reviewed the connection, buffering, finalize, protocol-mapping, packaging, and test paths. The offline suite passes (39 tests), and Black/Pylint are clean. I found three issues that should be addressed before merge.

kimdwkimdw:
Addressed the three review findings:

1. **Finalize boundary** — Fixed in b94f23052. Completion now checks the final result's end timestamp and pending hypothesis state instead of treating all bytes sent as finalized. Added regression coverage for delayed automatic finals followed by trailing audio.

2. **WebSocket 401 recovery** — Fixed in 431594fc5. A WebSocket handshake 401 now invalidates the token and retries once with a freshly issued token. Authentication-endpoint failures and repeated handshake 401s remain terminal. Added tests for revoked-token recovery and bounded retries.

3. **English model** — Fixed in 431594fc5. Removed the previous English model from the implementation, test configurations, and documentation. English now uses `model_name=whisper` with `language=en`, and reports `en-US` in TEN results.

diyuyi-agora (dismissed):

kimdwkimdw:
@diyuyi-agora Added small fix for GitHub CI.

https://github.com/TEN-framework/ten-framework/actions/runs/34855227460/job/104235631743?pr=2320

kimdwkimdw:
rebased from latest main branch

diyuyi-agora (approved):

